PR #605, opened 2026-04-24
chore: replace pytz with stdlib zoneinfo

### LINKED ISSUE #602: Bug: Datetime plugin seems broken - saving plugin config returns error 400
Image

Image

roblesi:
pytz was causing intermittent errors (see #602). Python 3.9+ ships `zoneinfo` (IANA time zone support) in the standard library, so we can drop the third-party dependency entirely.

Replacements applied across src, plugins, tests, and docs:
- `pytz.UTC` / `pytz.timezone("UTC")` -> `datetime.timezone.utc`
- `pytz.timezone(name)` -> `zoneinfo.ZoneInfo(name)`
- `pytz.exceptions.UnknownTimeZoneError` -> `(ZoneInfoNotFoundError, ValueError)`
- `tz.localize(naive_dt)` -> `naive_dt.replace(tzinfo=tz)` (zoneinfo is DST-aware natively and has no `.localize()`)

Also removes `pytz` from requirements.txt and pyproject.toml, and cleans up an unused `import pytz` in scripts/generate_sun_art_images.py.

Verified: full pytest suite green for all touched modules (time_service, date_time, countdown, integration + schedule tests). The 3 pre-existing failures in plugins/_template are unrelated (missing `get_config` from the V3 plugin base refactor).

Fixes #602

### DIFF
--- a/docs-site/docs/development/plugin-guide.md
+++ b/docs-site/docs/development/plugin-guide.md
@@ -442,7 +442,7 @@ from typing import Any, Dict, List, Optional
 import logging
 from datetime import datetime
 
-import pytz
+from zoneinfo import ZoneInfo, ZoneInfoNotFoundError
 
 from src.plugins.base import PluginBase, PluginResult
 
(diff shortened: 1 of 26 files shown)


danstaley:
This worked!

PR #607, opened 2026-04-25
fix: resolve pointer-events and scroll issues for dropdowns inside Sheet sidebars

jeffredodd:
## Problem

Several dropdown components were broken when used inside Sheet (slide-out sidebar) components:

1. **Select dropdowns were unclickable** — Radix Dialog/Sheet sets `document.body { pointer-events: none }` when open. Portaled Select dropdowns inherit this and swallow all mouse clicks. Keyboard navigation still worked since key events go directly to focused elements.

2. **TimezonePicker dropdown was unclickable** — Same body pointer-events issue. The custom `createPortal` dropdown rendered in `document.body` also had no mouse events.

3. **TimezonePicker dropdown mis-positioned on scroll** — When the Sheet content scrolled, the fixed-position portal dropdown stayed at its original coordinates rather than tracking the trigger element.

4. **TimezonePicker dropdown unscrollable via mouse wheel** — Radix's `react-remove-scroll` (used by Sheet) adds a document-level `wheel` listener that calls `preventDefault()` on events outside the modal DOM tree. Since the portal is in `document.body` (outside Sheet's React tree), wheel events were blocked. Dragging the scrollbar still worked.

### Radix Select inside Sheet
- Add `style={{ pointerEvents: "auto" }}` to `SelectPrimitive.Content` in the shared `select.tsx` component — directly overrides the body's rule for all portaled Select dropdowns
- Add `modal={false}` to Select components in `schedule-entry-form.tsx`, `carousels/page.tsx`, and `plugin-settings/schema-form.tsx` to prevent `DismissableLayer` from fighting over pointer-events

### TimezonePicker custom portal
- Add `pointerEvents: "auto"` to the portal `dropdownStyle` so mouse clicks register
- Add a capture-phase `scroll` listener to reposition the dropdown when any ancestor scrolls (e.g. Sheet), skipping repositioning when the scroll originates inside the dropdown itself
- Add a `wheel` event `stopPropagation()` listener on the portal element to prevent `react-remove-scroll` from blocking mouse wheel scroll inside the dropdown list

## Tests

Added `web/src/__tests__/schedule-entry-form-select.test.tsx` with 6 tests covering both mouse-click and keyboard interaction for all three Select components (page picker, start time, end time) inside a Sheet.

All 678 existing tests continue to pass (48 test files).

### DIFF
--- a/web/src/__tests__/schedule-entry-form-select.test.tsx
+++ b/web/src/__tests__/schedule-entry-form-select.test.tsx
@@ -1,0 +1,223 @@
+/**
+ * Tests for Select dropdown interactions (mouse and keyboard) inside a Sheet.
+ *
+ * Root cause of the original bug: Radix UI Dialog (Sheet) sets
+ * `document.body.style.pointerEvents = "none"` in modal mode. Radix Select
+ * portals its content into document.body, which then inherits that rule —
+ * so mouse clicks on Select options are silently swallowed. Keyboard events
+ * still work because they dispatch directly to the focused element.
+ *
+ * Fix: `modal={false}` on each Select nested inside a Sheet. The Sheet
+ * already handles modal behavior; the Select doesn't need to participate.
+ *
+ * These tests render ScheduleEntryForm inside an open Sheet to replicate
+ * the real usage context and assert that both mouse-click and keyboard
+ * interactions work correctly.
+ */
+
+import { describe, it, expect, vi } from "vitest";
+import { render, screen, waitFor } from "@testing-library/react";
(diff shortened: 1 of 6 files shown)


jeffredodd:
Closing this PR — the fix was already squash-merged into `main` as part of PR #609 (commit f9fffc3), which bundled the pointer-events/scroll fix together with the CI sentinel job. All 6 changed files (`select.tsx`, `schedule-entry-form.tsx`, `carousels/page.tsx`, `schema-form.tsx`, `timezone-picker.tsx`, and the new test file) are already on `main`.

PR #621, opened 2026-04-25
Potential fix for code scanning alert no. 25: Uncontrolled data used in path expression

jeffredodd:
Potential fix for [https://github.com/Fiestaboard/FiestaBoard/security/code-scanning/25](https://github.com/Fiestaboard/FiestaBoard/security/code-scanning/25)

The best fix is to enforce a **canonical path containment check inside `clone_or_update_repo`** before any filesystem use of `dest_dir`.  
That means:

- Normalize/resolve `dest_dir` with `os.path.realpath`.
- Resolve the trusted external root (`get_external_plugins_dir()`), also with `realpath`.
- Use `os.path.commonpath([external_root, candidate]) == external_root` to ensure `dest_dir` is inside the allowed directory.
- If not contained, return `(False, "...outside external plugins directory...")` before `dest_dir.exists()` or any git operations.

This is the single best non-breaking hardening because it preserves existing behavior for valid paths, adds defense-in-depth at the sink, and addresses all alert variants in one place.  
Edit only `src/plugins/sources.py` in `clone_or_update_repo(...)` near the start of the function (just after `env = ...` and before line 293 path usage). No new dependencies are needed.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/src/plugins/sources.py
+++ b/src/plugins/sources.py
@@ -290,6 +290,20 @@ def clone_or_update_repo(
     """
     env = {**os.environ, "GIT_TERMINAL_PROMPT": "0"}
 
+    # Defensive sink-level guard: ensure destination stays inside the
+    # managed external plugins directory before any filesystem access.
+    external_root = os.path.realpath(str(get_external_plugins_dir()))
+    candidate = os.path.realpath(str(dest_dir))
+    try:
+        if os.path.commonpath([external_root, candidate]) != external_root:
+            return False, (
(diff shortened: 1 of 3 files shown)


jeffredodd:
@copilot can you continue the fix here and get CI passing and make sure it validates the original intent too?

jeffredodd:
@copilot can you fix the ci issues here

PR #659, opened 2026-04-30
ci(fiestapi): unify all builds to attach to existing app release

jeffredodd:
Follow-up to #658.

After #658 merged we still have two FiestaPi-specific publish paths in `build-fiestapi.yml`:

1. **Major-version tag builds** (`vN.0.0`) — create their own release with a FiestaPi-specific body (Wi-Fi/flashing instructions).
2. **Weekly + manual builds** — create a separate `fiestapi-<version>` prerelease.

Both are duplicative. Each app version's release page (created by `release.yml`) should be the single home for both the code release and the matching FiestaPi `.img.xz`.

## Changes

- **Single `Resolve target app release tag` step** covers both paths:
  - Tag builds → use the pushed tag (`refs/tags/vN.0.0`).
  - Weekly cron / manual dispatch → resolve the latest published non-prerelease `v*` release via `gh release list`.
- **Single `Attach FiestaPi image to app release` step** uploads `.img.xz` and `.sha256` to the resolved tag. Body omitted so the existing app release notes are preserved. `softprops/action-gh-release@v2` replaces same-named assets in place, so reruns overwrite cleanly.
- **Checksum generation is now unconditional** (was tag-only) since every path needs it.
- **Drops the FiestaPi-only release body.** Wi-Fi/flashing instructions belong in `docs/setup/RASPBERRY_PI.md`, not duplicated per release.

Net diff: -57/+31 lines.

## Test plan

After merge, trigger `Build FiestaPi image` via `workflow_dispatch`. Verify:
- The `.img.xz` + `.sha256` show up as assets on the latest app release (currently https://github.com/Fiestaboard/FiestaBoard/releases/tag/v5.1.1).
- Release body is unchanged.
- Re-triggering replaces assets in place rather than appending duplicates.
- The orphan https://github.com/Fiestaboard/FiestaBoard/releases/tag/fiestapi-2026-W18 prerelease can be deleted manually — won't recur.

## Note on tag-build race

On a `vN.0.0` push, `release.yml` and `build-fiestapi.yml` both touch the same release. `softprops` creates if missing / updates if present, and we no longer write a body in `build-fiestapi.yml`, so there's no clobber risk regardless of order.


jeffredodd:
@copilot+claude-opus-4.7 can you resolve conflicts here?

PR #711, opened 2026-05-07
feat: Add "Quiet Library" batched word-by-word transition

wonkybutt:
> **Maintainer note:** This PR has been rebuilt on top of the transition-plugin SDK that landed in #829. The original `quietLibrary` strategy (hardcoded in `board_client.py`) is now a self-contained plugin at `plugins/quiet_library/` — same algorithm, same acoustic profile, credit to @wonkybutt. The diff is just the plugin, its README row, and a one-line CI guard.

## What's in the plugin

* **Word Boundary Isolation** — scans the grid and groups contiguous modified characters (and trailing spaces) into isolated word blocks. Now W×H-aware via `BoardContext`, so it works on note arrays too, not just the 6×22 flagship.
* **Micro-Batching (max 6 tiles)** — modifications within a word block are subdivided into batches of ≤6 tiles; a single batch never spans two words.
* **Hardware Debounce Bypass (14.5s default step delay)** — a strict per-frame delay clears the Vestaboard's internal flap debounce so no batch is dropped mid-rotation. Configurable via `step_delay_ms`.
* **Runner-managed execution** — background threading, interruption, rate-limit pacing, and the final snap-to-target are handled by the platform's `TransitionRunner` (#829), so the plugin is pure frame generation.
* **Tests** — 20 tests, 100% coverage (`python scripts/run_plugin_tests.py --plugin=quiet_library`).
* **Docs** — plugin README, SETUP guide, and board screenshot in `plugins/quiet_library/`.
* **CI guard (drive-by fix)** — `release.yml` now skips on forks (`if: github.repository == 'Fiestaboard/FiestaBoard'`). Without it, any fork that pushes to its `main` runs the release pipeline and self-bumps its version files — which is exactly what polluted this PR's branch with a stray `8.8.2` bump commit.

## Original description (@wonkybutt)

This pull request introduces the `quietLibrary` transition strategy, designed to update the board with the minimum possible mechanical noise. The behavior is modeled after the split-flap board in the Qantas First Class lounge at the Sydney airport, where the display updates methodically one word at a time to maintain a quiet environment.

To achieve this specific acoustic profile on the Vestaboard hardware, a custom row-by-row, word-by-word diffing algorithm was implemented.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [X] New plugin
- [ ] Documentation update
- [ ] Refactor / maintenance

## Checklist

- [X] I have tested my changes locally
- [X] I have added/updated tests where applicable
- [X] I have updated documentation where applicable
- [X] My code follows the project's coding standards

Follow-up to #829 (merged) — this lands the fourth first-party transition plugin.


jeffredodd:
Hey @wonkybutt thanks for the contribution! I think this is a really great idea but am also trying to think of the best way to implement this. The transitions you're trying to extend relate directly to the Vestaboard's local API and parameters it receives when we set them to various things like row transition. So because these classes are meant to mirror what Vestaboard provides I'm not sure it's the right spot for this to live long term. 

But what I'm thinking is there is a need for some filter/animation/sequencer (Not sure on a name yet) that allows you to do just what you're trying to do, provide a frame by frame update, at a given interval, to produce an effect. 

Anyways, give me a bit to noodle on this and I might want to build upon this idea if thats ok with you. Thanks again!

roblesi:
> Hey @wonkybutt thanks for the contribution! I think this is a really great idea but am also trying to think of the best way to implement this. The transitions you're trying to extend relate directly to the Vestaboard's local API and parameters it receives when we set them to various things like row transition. So because these classes are meant to mirror what Vestaboard provides I'm not sure it's the right spot for this to live long term.
> 
> But what I'm thinking is there is a need for some filter/animation/sequencer (Not sure on a name yet) that allows you to do just what you're trying to do, provide a frame by frame update, at a given interval, to produce an effect.
> 
> Anyways, give me a bit to noodle on this and I might want to build upon this idea if thats ok with you. Thanks again!

+1

jeffredodd:
@wonkybutt just an update here. We're going to expose transitions that can be extended via plugin. The idea here is that you can write plugins that will transition your Vestaboard from one state to another. That puts the keys in your hands for how to handle updates coming in during a transition, as well as other config settings for your transition since I think everyone is going to want to tweak things per plugin. 

Anyways stay tuned it's on our roadmap.

wonkybutt:
@jeffredodd Oh awesome! I was just starting to think about sending specific transitions based on different content. Looking forward to see what's possible!

jeffredodd:
Hi @wonkybutt — maintainer update on this PR. 👋

Since you opened this, the transition system was rebuilt as a **plugin SDK** (#829): frame-by-frame transitions are now self-contained plugins instead of strategies hardcoded in `board_client.py`. Rather than closing your PR, we've rebuilt it on that SDK so it merges under your name:

- Your branch now carries #829's SDK plus one commit adding `plugins/quiet_library/` — your algorithm (word-boundary diffing, ≤6-tile micro-batches, 14.5s debounce-clearing step delay) as a plugin, co-authored to you.
- Threading, interruption, and rate-limit pacing are handled by the platform's `TransitionRunner` now, so the plugin is pure frame generation — and it works on note arrays too.
- Once #829 merges, this PR's diff reduces to just the plugin, and we'll merge it right after.

Heads up that this required a force-push to your fork's `main` (the PR head). If you have local work there, rebase on the new tip before pushing. Thanks again for the concept and the algorithm — the Qantas-lounge acoustic profile survives intact, with 20 tests at 100% coverage.

PR #712, opened 2026-05-07
Page editor: show active state on wrap and alignment toolbar buttons

jeffredodd:
## Summary

- Wire `currentAlignment` through to the TipTap toolbar (it was destructured as `_currentAlignment` and silently dropped), so the left/center/right buttons reflect the current line's alignment. The "left" button now lights up on default lines, since `currentAlignment` defaults to `'left'`.
- Strengthen the wrap button's active treatment to match — it had an active style but the subtle `bg-muted/70` + recolored icon read as inactive next to the unstyled alignment buttons.
- Both controls now use `bg-primary text-primary-foreground` when active, matching the fallback alignment controls already rendered below the editor.

No data-model, prop, or upstream changes. Only `TemplateEditorToolbar.tsx` is touched.

## Test plan

- [ ] Open the Page editor on a fresh line — the **left** alignment button is active.
- [ ] Center- or right-align the line — the corresponding button activates and the others deactivate; cursor moves between lines update the active button.
- [ ] Toggle wrap on a line — the wrap button shows the prominent active state; moving the cursor to a non-wrapped line returns it to inactive.
- [ ] Multi-line selection still applies alignment to the full range.
- [ ] Save and reload — active states reflect the persisted per-line metadata.

### DIFF
--- a/web/src/components/tiptap-template-editor/components/TemplateEditorToolbar.tsx
+++ b/web/src/components/tiptap-template-editor/components/TemplateEditorToolbar.tsx
@@ -35,7 +35,7 @@ interface TemplateEditorToolbarProps {
 
 export function TemplateEditorToolbar({
   editor,
-  _currentAlignment = 'left',
+  currentAlignment = 'left',
   currentWrapEnabled = false,
   onAlignmentChange,
   onWrapToggle,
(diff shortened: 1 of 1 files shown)


jeffredodd:
@copilot+claude-opus-4.7 Can you fix the tests here so our CI passes?

jeffredodd:
@copilot+claude-opus-4.7 can you fix the CI issues here?

PR #818, opened 2026-05-30
fix(pi): self-heal first-boot pull race + rename image output

jeffredodd:
## Summary

Two related fixes for the FiestaPi image:

- **First-boot pull race**: on fresh SD cards, `docker compose up -d` pulls `fiestaboard` and `fiestaupdater` in parallel, which races containerd's snapshot extraction (`lease does not exist: not found`). With `Type=oneshot` and no `Restart=` policy, the unit dies once and never retries — port 4420 stays closed and the user sees "no login response on a fresh install." Reproduced live on a Pi 4 over Ethernet during this session.
- **Image filename rename**: pi-gen's default output `image_<build-date>-fiestapi-lite.img.xz` is ugly, missing the FiestaBoard version, and inconsistent with the docs (which have always told users to download `fiestapi-<version>-arm64.img.xz`). Rename to `FiestaPi-<version>-arm64.img.xz` and bring all current docs in line.

## Changes

**`pi-image/.../fiestaboard.service`**

- `Environment=COMPOSE_PARALLEL_LIMIT=1` — serializes layer extraction so the containerd snapshot race cannot occur on slow SD-card I/O.
- `ExecStartPre=-/usr/bin/docker compose pull` — explicit pre-pull so transient failures show up in the journal and are counted against the retry budget. The `-` prefix tolerates offline boots.
- `Restart=on-failure`, `RestartSec=30s`, `StartLimitIntervalSec=600`, `StartLimitBurst=10` — any transient pull/extract failure self-heals within ~5 minutes instead of stranding the user.

**`.github/workflows/build-fiestapi.yml`**

- Rename pi-gen's output to `FiestaPi-<version>-arm64.img.xz` before checksum/upload/release-attach. Workflow artifact name follows. Tag builds (`v6.1.1`) yield `FiestaPi-6.1.1-arm64.img.xz`; weekly cron yields `FiestaPi-2026-WNN-arm64.img.xz`.

**Docs** (current versions only — versioned snapshots left as historical record)

- `README.md`, `pi-image/README.md`, `docs/setup/RASPBERRY_PI.md`, `docs-site/docs/setup/{raspberry-pi,beginners-guide}.md` updated to reference `FiestaPi-<version>-arm64.img.xz`.

## Why parallel pulls race

`pull_policy: always` on both services in `pi-image/.../docker-compose.yml` plus `COMPOSE_PROFILES=fiestaupdater` in the .env means every `docker compose up` pulls two images. On a fresh SD card with no cached layers, containerd extracts both image trees concurrently, and intermittently loses track of a lease during snapshot commit. The error surfaces as `failed to commit snapshot extract-<id>-<rnd> sha256:<digest>: NotFound: lease does not exist: not found`. Plenty of disk available (108 GB free on the test Pi) — it's a concurrency bug, not a resource one.

## Recovering existing customers (no reflash)

Already-deployed Pis that hit this can recover by SSHing in and restarting the unit. Once the images are cached locally, subsequent `up` invocations skip the racy pull path:

```
ssh fiesta@fiestapi.local   # password: fiestaboard
sudo systemctl restart fiestaboard
```

## Test plan

- [x] Reproduced the failure live on a fresh Pi 4 (Ethernet, 192.168.0.154) — `docker ps -a` empty, `journalctl -u fiestaboard` shows the containerd lease error, port 4420 closed.
- [x] Deployed the patched unit to the test Pi via scp + `daemon-reload` + manual pull/restart — port 4420 opens, `/api/auth/status` returns `{"setup_required":true,"first_run":true}`.
- [x] Confirmed the unit reloads cleanly with `systemctl restart` after the patch.
- [ ] CI build of `FiestaPi-<version>-arm64.img.xz` succeeds (in progress: https://github.com/Fiestaboard/FiestaBoard/actions/runs/26689164974).
- [ ] Flash the resulting image to a fresh SD card and confirm first boot reaches `setup_required:true` over both Ethernet and Wi-Fi.

## Follow-ups (not in this PR)

- **Pre-pull images during pi-gen build** (largest UX win): use the arm64 CI runner to `docker save` both images into the rootfs and `docker load` them in `firstboot.sh` before `up -d`. Eliminates the first-boot network dependency entirely and drops ~3 min off first boot. Image grows ~600 MB compressed.
- **`docker compose up -d --wait`**: make systemd `active` mean "healthcheck green" instead of "compose returned 0". Honest state for downstream units, ~1–2 min later activation.
- **Verbose `/dev/console` output from `firstboot.sh`** for users debugging with a monitor attached.

### DIFF
--- a/.github/workflows/build-fiestapi.yml
+++ b/.github/workflows/build-fiestapi.yml
@@ -182,9 +182,14 @@ jobs:
           retention-days: 30
 
       - name: Resolve target app release tag
-        # Every build path attaches to an existing app release:
+        # Skip on workflow_dispatch from non-default branches: a manual
+        # build from a feature branch should produce a verifiable artifact
+        # for testing, not pollute a published release page. Dispatch on
+        # main remains attached (maintainer hot-patch path).
+        if: github.event_name != 'workflow_dispatch' || github.ref == 'refs/heads/main'
+        # Every release-attaching build path resolves an existing app release:
         #   - Tag builds (refs/tags/vN.0.0) → attach to that tag's release.
-        #   - Weekly cron / manual dispatch → attach to the latest published
+        #   - Weekly cron / dispatch-on-main → attach to the latest published
         #     non-prerelease app release (e.g. v5.1.1).
         # This keeps each app version's release page as the single source
         # of truth for both the code release and its FiestaPi .img.xz.
@@ -207,6 +212,7 @@ jobs:
           echo "Will attach FiestaPi image to release: $TAG"
 
       - name: Attach FiestaPi image to app release
+        if: github.event_name != 'workflow_dispatch' || github.ref == 'refs/heads/main'
         # Upload the .img.xz and .sha256 to the resolved release.
         # softprops/action-gh-release@v3 replaces same-named assets in
         # place, so re-running overwrites cleanly. Body is intentionally
--- a/.github/workflows/build-fiestapi.yml
+++ b/.github/workflows/build-fiestapi.yml
@@ -118,7 +118,7 @@ jobs:
             fi
           else
             # Weekly builds labelled by year-week so the filename is
-            # meaningful (e.g. fiestapi-2026-W18-arm64.img.xz).
+            # meaningful (e.g. FiestaPi-2026-W18-arm64.img.xz).
             VERSION="$(date -u +%Y-W%V)"
           fi
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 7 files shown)


jeffredodd:
## Self-review pass

Two findings worth surfacing:

**1. Caught and fixed: feature-branch dispatches were polluting public releases** (now commit `5d54399`)

The existing `Attach FiestaPi image to app release` step ran unconditionally. The first build I dispatched from this branch would have uploaded `FiestaPi-2026-W22-arm64.img.xz` to the public **v6.1.1** release page in ~45 minutes, where customers would see it alongside the real release asset. Cancelled it (`gh run cancel 26689164974`) and added an `if:` guard on both `Resolve target app release tag` and the attach step:

```yaml
if: github.event_name != 'workflow_dispatch' || github.ref == 'refs/heads/main'
```

Preserves the maintainer's documented intent (dispatch-on-main attaches; weekly cron attaches; tag builds attach) while making feature-branch dispatches artifact-only. Re-triggered build: https://github.com/Fiestaboard/FiestaBoard/actions/runs/26689351100.

**2. Confirmed live: `Restart=on-failure` works with `Type=oneshot` on Debian trixie / systemd 256**

I deployed the patched unit to the test Pi and installed a drop-in that overrode `ExecStart=/bin/false` to force failure. Journal showed:

```
16:44:01 fiestaboard.service: Main process exited, code=exited, status=1/FAILURE
16:44:01 fiestaboard.service: Failed with result 'exit-code'.
16:44:11 fiestaboard.service: Scheduled restart job, restart counter is at 1.
```

Exactly the 10s `RestartSec` I set in the drop-in. Removed the drop-in, restored normal operation, both containers back up.

## Items I considered but left alone

- **Versioned doc snapshots** (`docs-site/versioned_docs/*`) — these describe past releases where the real filename was `image_<date>-fiestapi-lite.img.xz`; touching them to match the new naming would misrepresent history. Forward-looking rename only.
- **`docker compose up -d --wait`** — would make systemd `active` honest about healthcheck state, but adds 1–2 min to startup. Deferred to follow-up; this PR is already an effective fix.
- **Pre-pull images during pi-gen build** — the biggest UX win (eliminates first-boot network dependency entirely), but a non-trivial pipeline change; deferred to its own PR.

PR #845, opened 2026-05-31
refactor(mcp): return structured dicts instead of JSON-string output

jeffredodd:
## Summary

Stacked on **#837** (which is stacked on **#835**). Final piece of the MCP-quality work.

Every tool was wrapping its return in \`json.dumps(...)\` and declaring \`-> str\`. MCP clients then had to parse the string back to JSON, and the tool schema surfaced as a plain text blob rather than the structured shape the tool actually produces.

This refactor makes every tool return native \`dict\` / \`list\` values:
- **FastMCP serializes the return value natively** → clients see real JSON, and the MCP tool schema reflects the actual return shape.
- **Successful mutations return a uniform envelope**: \`{\"status\": \"success\", \"message\": \"...\", **context_fields}\`
- **Errors return**: \`{\"status\": \"error\", \"error\": \"...\"}\`
- **Read tools return the underlying object** (list of pages, plugin dict, etc.) — no envelope needed when the shape is already structured.

Three small helpers (\`_ok\`, \`_err\`, \`_serialize\`) keep the patterns consistent. \`_serialize\` handles Pydantic models (\`.model_dump(mode=\"json\")\`), dataclasses (\`dataclasses.asdict\`), datetimes, and ad-hoc objects so callers don't have to think about JSON-encoding edge cases.

## Test plan

- [x] \`pytest tests/test_mcp_server.py\` — 91 passed
- [x] Full suite \`pytest tests/\` — 3571 passed, 11 skipped
- [x] Test assertions updated to consume dict/list directly (\`assert result[\"status\"] == \"success\"\` etc.) instead of re-parsing JSON strings
- [x] Error-resilience parametrize test asserts \`isinstance(result, (dict, list))\` — confirms every tool returns structured data even when every backing service raises

### DIFF
--- a/src/mcp_server.py
+++ b/src/mcp_server.py
@@ -25,11 +25,55 @@ Connection example for Claude Code:
 
 from __future__ import annotations
 
-import json
 import logging
 from typing import Any
 
 logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Response helpers — every tool returns structured data (dict/list) rather
+# than a json.dumps()'d string. FastMCP serializes the return value into
+# tool output automatically, so clients get real JSON instead of a JSON
+# string that has to be parsed again.
+# ---------------------------------------------------------------------------
+
+def _ok(message: str, **fields: Any) -> dict[str, Any]:
+    """Standard success envelope for mutation tools."""
+    return {"status": "success", "message": message, **fields}
+
+
+def _err(error: str) -> dict[str, Any]:
+    """Standard error envelope. Tools never raise — they return this instead."""
+    return {"status": "error", "error": error}
+
+
+def _serialize(obj: Any) -> Any:
+    """Convert Pydantic models / dataclasses / datetimes to JSON-compatible primitives.
+
+    Used when a tool returns Pydantic models (pages, schedules, carousels)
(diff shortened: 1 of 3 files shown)


jeffredodd:
Note from cleanup: when rebasing this branch onto current main (which now includes #848's HTML-preview embedding via `_page_response_with_preview`), `src/mcp_server.py` produces real design-level conflicts in `get_page`, `create_page`, and related tools — 848 returns previews embedded in responses, while this PR refactors the same tools to return structured dicts via `_serialize`/`_ok`/`_err` helpers. The two approaches need to be reconciled before this can merge cleanly (probably: have the new helpers also embed the preview when applicable). Not auto-resolvable.

jeffredodd:
Rebased onto current main. Reconciliation note: this PR's structured-dict refactor conflicts at design level with #836's `_page_response_with_preview` HTML embedding in `get_page`/`create_page`/`update_page`. Resolved by taking the structured-dict design here and removing the embed-on-page-tool behavior — the dedicated `render_page_preview` tool (from #837) still provides preview functionality on demand. `_page_response_with_preview` helper and the corresponding `TestPageToolsEmbedHtmlPreview` test class are removed. 3640 tests pass locally; ruff is clean.

PR #909, opened 2026-06-07
feat(nginx): make X-Frame-Options / frame-ancestors configurable via env

jeffredodd:
## Why

Hard-coding `add_header X-Frame-Options "SAMEORIGIN" always;` in nginx prevents FiestaBoard from being embedded in iframes when the embedding context is treated as cross-origin by the browser — most notably the Home Assistant add-on sidebar, where [HA Ingress renders the add-on UI inside a sandboxed iframe](https://developers.home-assistant.io/docs/add-ons/presentation#ingress). Browsers treat sandboxed frames as having an opaque origin, so `SAMEORIGIN` denies framing even though the iframe's URL is on the same host as the parent page.

We hit this from the downstream Home Assistant add-on at [Fiestaboard/FiestaBoard-Home-Assistant-App](https://github.com/Fiestaboard/FiestaBoard-Home-Assistant-App): the add-on loads fine in a fresh tab but the sidebar panel is blank.

## What

Move the frame-embedding directive into `/etc/nginx/snippets/frame-headers.conf`, generated at boot by a new `configure_frame_headers()` in `entrypoint.sh` from two new env vars:

| Env var | Default | Notes |
| --- | --- | --- |
| `FIESTABOARD_X_FRAME_OPTIONS` | `SAMEORIGIN` | Historical behavior. Values like `DENY` are emitted as-is. Special value `OFF` suppresses the header so a CSP `frame-ancestors` directive can fully control framing in modern browsers. |
| `FIESTABOARD_FRAME_ANCESTORS` | (unset) | When set, emits `Content-Security-Policy: frame-ancestors <value>`. Examples: `'self'`, `'self' https://my.host`, `*`. |

`nginx.conf`, `nginx.https.conf`, and `nginx-dev.conf` use `include /etc/nginx/snippets/*.conf;` (wildcard, so a missing snippet directory does not error during local dev where the entrypoint may be bypassed).

## Backward compatibility

When neither env var is set the rendered snippet contains exactly the same `add_header X-Frame-Options "SAMEORIGIN" always;` line that was there before, so existing deployments see no observable change in HTTP headers.

Smoke-tested all three branches of `configure_frame_headers` in a local shell:

| Env | Rendered snippet |
| --- | --- |
| _(unset)_ | `add_header X-Frame-Options "SAMEORIGIN" always;` |
| `FIESTABOARD_X_FRAME_OPTIONS=OFF FIESTABOARD_FRAME_ANCESTORS="'self'"` | `add_header Content-Security-Policy "frame-ancestors 'self'" always;` |
| `FIESTABOARD_X_FRAME_OPTIONS=DENY` | `add_header X-Frame-Options "DENY" always;` |

## Test plan

- [ ] Build the image with no env overrides; `curl -sI http://localhost:3000/` returns `X-Frame-Options: SAMEORIGIN` (no change vs. main).
- [ ] Rebuild with `FIESTABOARD_X_FRAME_OPTIONS=OFF` + `FIESTABOARD_FRAME_ANCESTORS="'self'"`; `curl -sI` returns no `X-Frame-Options` and `Content-Security-Policy: frame-ancestors 'self'`.
- [ ] Verify the downstream HA add-on (sibling PR in [FiestaBoard-Home-Assistant-App](https://github.com/Fiestaboard/FiestaBoard-Home-Assistant-App)) renders the FiestaBoard panel inside the HA sidebar iframe.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -115,7 +115,8 @@ RUN mkdir -p /app/static
 COPY starting.html /app/static/starting.html
 
 # Create nginx directories and set permissions
-RUN mkdir -p /var/log/nginx /var/lib/nginx/tmp /run/nginx /var/lib/nginx/body
+# /etc/nginx/fiestaboard is our dedicated snippet dir (avoids Debian's /etc/nginx/snippets/).
+RUN mkdir -p /var/log/nginx /var/lib/nginx/tmp /run/nginx /var/lib/nginx/body /etc/nginx/fiestaboard
 
 # Create data directory for logs and app state
 RUN mkdir -p /app/data/logs
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -91,8 +91,57 @@ except (FileNotFoundError, json.JSONDecodeError, OSError):
     fi
 }
 
+# ---------------------------------------------------------------------------
+# Frame-embedding headers
+# ---------------------------------------------------------------------------
+# nginx.conf / nginx.https.conf / nginx-dev.conf each `include` the snippets
+# directory inside their `server` block; this function renders the actual
+# `add_header` lines into /etc/nginx/snippets/frame-headers.conf so they can
+# be controlled by env without baking a value into the image.
+#
+# Env vars (both optional):
+#   FIESTABOARD_X_FRAME_OPTIONS  Defaults to "SAMEORIGIN" (historical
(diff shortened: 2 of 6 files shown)


jeffredodd:
@claude Can you fix the CI tests and get them passing?

jeffredodd:
@claude We have another e2e failure, can you fix it?

PR #913, opened 2026-06-07
fix(nginx): honor X-Ingress-Path so Next.js loads behind HA Ingress

jeffredodd:
## Summary
- HA Supervisor's Ingress wraps add-on UIs in an iframe at `/api/hassio_ingress/<token>/` and signals the prefix via `X-Ingress-Path`. FiestaBoard's HTML currently emits absolute `/_next/...` and `/api/...` paths, so the browser fetches them from the host's *origin root*, bypasses the proxy, and gets a 404 from HA core — visible as `Refused to execute … nosniff` console errors and a blank/broken sidebar iframe.
- In the `location /` block of `nginx.conf` / `nginx.https.conf` / `nginx-dev.conf`, add four `sub_filter` rules that prepend `$http_x_ingress_path` to absolute `/_next/` and `/api/` references in HTML responses (both `"` and `'` quote styles). Upstream gzip is disabled with `proxy_set_header Accept-Encoding ""` so sub_filter can see plaintext; the outer `gzip on` directive re-compresses the rewritten response before it leaves nginx.
- When the header is absent, every replacement degenerates to a self-substitution, so direct (non-Ingress) access and standalone Docker installs are byte-for-byte unaffected.

## Companion change
Once this lands and a release ships, the [FiestaBoard-Home-Assistant-App add-on](https://github.com/Fiestaboard/FiestaBoard-Home-Assistant-App) will bump `BUILD_FROM` and (no shim changes needed — HA Supervisor already sends `X-Ingress-Path` automatically).

## Scope / limitations
This PR fixes the **initial HTML chunk 404s** — the immediate user-visible symptom. Next.js client-side router navigation (Link components, `router.push`) still uses absolute paths baked at build time and will need a follow-up — either a runtime `assetPrefix` plumbed through `next.config.ts`, or a small piece of client code that reads the prefix from `window.location.pathname` and sets `__webpack_public_path__` early enough.

## Test plan
- [x] `nginx -t` passes for all three configs (verified inside `nginx:alpine`).
- [x] Functional test in `nginx:alpine` with a Python upstream serving sample Next.js HTML:
  - Without `X-Ingress-Path`: response bytes identical to upstream.
  - With `X-Ingress-Path: /api/hassio_ingress/ABC123`: `/_next/...` and `/api/...` references rewritten to `/api/hassio_ingress/ABC123/_next/...` and `/api/hassio_ingress/ABC123/api/...`. `/dashboard` (out of scope) left untouched.
- [ ] Test against a real HA Supervisor install with the add-on bumped to a build containing this image (will validate in companion PR).

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -139,9 +139,87 @@ configure_frame_headers() {
     fi
 }
 
+# ---------------------------------------------------------------------------
+# Reverse-proxy base-path rewriting (e.g. Home Assistant Ingress)
+# ---------------------------------------------------------------------------
+# Some embedding contexts (HA Supervisor Ingress, traefik subpath routing)
+# mount FiestaBoard under a per-installation URL prefix and signal that
+# prefix to the upstream via an `X-Ingress-Path` request header. Without
+# any rewriting, the browser sees `<script src="/_next/...">` in the
+# returned HTML, resolves it against the iframe's *origin root*, bypasses
+# the proxy, and 404s -- visible to users as "Refused to execute ...
+# nosniff" console errors and a broken sidebar iframe.
+#
+# When FIESTABOARD_INGRESS_PATH_REWRITE=true, render a `location /`
+# snippet that:
+#   1. Disables upstream gzip (sub_filter cannot rewrite compressed
+#      payloads; nginx's outer `gzip on` still compresses the rewritten
+#      response before it leaves the box).
+#   2. Adds sub_filter rules that prepend `$http_x_ingress_path` to
+#      absolute `/_next/` and `/api/` references in HTML responses,
+#      covering both `"` and `'` quote styles.
+#
+# When the env var is unset/false the snippet directory stays empty, so
+# direct (non-proxy) deployments incur zero buffering or gzip overhead.
+# Inside the snippet, when X-Ingress-Path is absent the substitutions
+# degenerate to self-substitutions, so toggling the var ON for a direct
+# deployment is also safe (just wasteful).
+#
+# Scope: this handles the initial HTML and its inline asset references,
+# which is what resolves the immediate chunk 404s seen behind HA Ingress.
+# Next.js client-side router navigation uses paths baked at build time
+# and is not yet covered here -- callers that need full SPA navigation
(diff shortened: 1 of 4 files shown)


jeffredodd:
Heads-up: I pushed a follow-up commit that addresses two issues:

1. **CI feedback.** First run hit two E2E failures (`login.redirecting`, `picks.tab-flagship`). Both snapshots showed the page successfully past the assertion point — symptoms consistent with timing/race flakes rather than a deterministic regression — but rather than just retry, I dug into the actual concern those failures raised.

2. **Architectural concern.** The original cut enabled `sub_filter` and disabled upstream gzip unconditionally on `location /`. That imposed real overhead on every deployment (sub_filter buffers responses → kills Next.js streaming SSR's first-byte latency, plus doubles bytes between Next.js and nginx) just to support proxy-prefix deployments. Standalone Docker / public-preview users shouldn't pay for a feature they don't use.

**What changed in 38e86c27**

Switched to the same opt-in pattern #909 used for frame headers:

- `entrypoint.sh` gets a `configure_ingress_path_rewrite()` function that renders the snippet into `/etc/nginx/fiestaboard/location-root/base-path-rewrite.conf` **only when `FIESTABOARD_INGRESS_PATH_REWRITE=true`**.
- `location /` in each nginx config now just `include`s that directory. When the env var is unset (the default), the include is a silent no-op and proxy behavior is byte-for-byte identical to pre-PR.
- Operators behind HA Supervisor Ingress (or any X-Ingress-Path-aware proxy) set the env var; everyone else gets the previous zero-overhead path.
- Toggling the var off across container restarts cleans up the stale snippet, so the modes converge.

**Verified**
- `nginx -t` passes on all three configs both with the snippet present and absent.
- Functional test in `nginx:alpine` confirms: with env var off → HTML byte-identical, with env var on + `X-Ingress-Path: /api/hassio_ingress/ABC` → all `/_next/` and `/api/` references rewritten as expected.
- Snippet lifecycle test confirms create-on-enable, remove-on-disable, remove-on-unset.

PR title is unchanged; this is the same feature with cleaner scoping. Should also remove the E2E timing pressure from the previous run by not touching the hot path for direct deployments.

Issue #941, opened 2026-06-11
Demo Page create for wrong Vestaboard size


JCoxRSOS:
**Describe the bug**
When a Demo Page is created, it seems to default to Flagship size. That size does not display well on the Note. 

**To Reproduce**
1. Configure FiestaBoard for Vestaboard Note
2. Install a integration such as 'Date & Time'
3. Create a Demo Page
4. Show the Demo Page on the Note
6. Observe that much of the content is cut off on 

**Expected behavior**
Integrations should include demo pages for both Vestaboard sizes, Flagship and Note. FiestaBoard should create a Demo Page based on the Vestaboard size configured in Settings > Hardware.

**Screenshots / Logs**
<img width="888" height="823" alt="Image" src="https://github.com/user-attachments/assets/56f838bb-0713-4161-9182-294da01de3d5" />

**Environment**
- OS: Ubuntu 22.04.4 LTS
- Docker version: 29.5.3
- FiestaBoard version: 7.10.0
- Board connection mode: Cloud
- Hardware: MFF PC

**Additional context**
N/A

JCoxRSOS:
Created user work GH account. Whoops. Please delete.

PR #952, opened 2026-06-11
fix(pages): show device-type tabs when orphan-device pages exist

### LINKED ISSUE #943: Edit/Delete Demo Pages
Yes, at present FiestaBoard allows plug-ins to create demo pages, but users don't have the ability to update or delete said demo pages.

The ability to update or delete demo pages, just like user-created pages.

I thought about finding them in the config files and editing them out manually, but I'd rather have a button.

**Additional context**
N/A

jeffredodd:
Fixes #943.

## Problem

When a plugin creates a demo page for a device type the user doesn't have configured (e.g. flagship demo pages on a note-only setup, see #941), those pages were completely hidden on `/pages`. The device-type tabs only appeared when `boardSettings.devices.length > 1`, so a user with only a note board never saw a "Flagship" tab — and the grid was filtered to note-only — so there was no way to edit or delete the orphan pages.

## Fix

Compute `availableDevices` as the union of the user's configured devices and the device types actually present in their pages. Tabs render whenever that union has more than one device, with the user's configured device selected by default. Orphan-device pages remain reachable via the secondary tab.

## Test plan

- [x] Unit tests in `web/src/__tests__/pages-index-device-tabs.test.tsx` cover four scenarios:
  - Tabs hidden when only the configured device's pages exist
  - Both tabs appear when an orphan flagship page exists on a note-only setup
  - The configured device tab is selected by default (not the orphan one)
  - Clicking the orphan flagship tab reveals the orphan page
- [x] Full vitest suite: **1044 passed**, 13 skipped
- [x] `tsc --noEmit` clean for the changed files
- [x] Manually reproduced the bug scenario in the running dev container (single note board + a flagship page via API) and confirmed both tabs appear on `/pages`, with the flagship page editable/deletable from its tab


jeffredodd:
@claude can you fix CI failures?

PR #955, opened 2026-06-11
feat(silence): tighten boundary race + show imminent-silence banner

jeffredodd:
## Summary
- **Race fix in `src/main.py`:** added a 1s-granularity silence-boundary detector on top of the existing `schedule.every(N).seconds.do(check_and_send_active_page)` poll. The silence page used to lag the configured boundary by up to `polling_interval` seconds (15s default, longer in cloud mode); now it kicks in within ~1s. No new threads, no config-change re-arming — the loop already runs `time.sleep(1)` between `run_pending()` calls, so adding `Config.is_silence_mode_active()` there is cheap (pure time math).
- **`SilenceImminentBanner` on the home page** for user-controlled remediation: shows when silence is enabled, `mode === "page"`, a `page_id` is set, the silence page isn't already showing, and silence starts within ≤2 min. CTA = "Switch now" → `POST /settings/temporary-override` with `revert_mode: "schedule"`. Dismissible (per upcoming boundary). When silence naturally starts, the normal silence flow takes over.
- **`/silence-status` now returns `seconds_until_next_change`** so the frontend doesn't re-do the UTC + midnight-rollover math the backend already does. `SilenceStatus` TS interface updated to include the new field and the existing `page_id`.

## Test plan
- [x] `pytest tests/test_silence_schedule_endpoint.py tests/test_silence_mode_display.py tests/test_silence_schedule_polling.py tests/test_silence_config_properties.py tests/test_time_service.py` → 127 passed (4 new under `TestSilenceStatusSecondsUntilNextChange`)
- [x] `vitest run src/__tests__/silence-imminent-banner.test.tsx` → 8/8 (renders within window, hides when disabled/active/far/mode-mismatch/already-showing-silence-page, dismiss works, Switch-now POSTs override with correct page id)
- [x] `vitest run src/__tests__/home-page-banner.test.tsx src/__tests__/active-page-display-silence.test.tsx` → no regressions
- [x] `eslint` clean on touched files; `ruff` clean on Python files
- [ ] Manual smoke: configure silence + mode=page + page_id, set start_time ~3 min in future, confirm banner appears at home; click Switch now → board flips to silence page immediately; verify silence kicks in at start_time and stays on that page

### DIFF
--- a/src/api_server.py
+++ b/src/api_server.py
@@ -3811,6 +3811,21 @@ async def get_silence_status():
     # Determine next change time (simplified - just return start or end)
     next_change_utc = end_time if active else start_time
 
+    # Wall-clock seconds until the next active/inactive transition. Lets the
+    # frontend show a "silence starts in N min" warning without re-doing the
+    # UTC + offset math the silence window uses (which has subtle edge cases
+    # around midnight rollover and DST). None when silence is disabled.
+    seconds_until_next_change: int | None = None
+    if enabled:
+        next_change_dt = time_service.parse_iso_time(next_change_utc)
+        if next_change_dt is not None:
+            delta_seconds = int((next_change_dt - current_utc).total_seconds())
+            # next_change_dt is anchored to "today" in UTC, so a negative value
+            # means the boundary already passed today and will recur tomorrow.
+            if delta_seconds < 0:
+                delta_seconds += 86_400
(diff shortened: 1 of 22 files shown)


jeffredodd:
@claude can you fix CI failures?

jeffredodd:
@claude the push should not have been blocked you have permissions

PR #957, opened 2026-06-11
chore(web): ignore stale Next.js artifacts

jeffredodd:
## Summary
- The web app migrated off Next.js to Vite + React Router v7, but `web/.next/` and `web/next-env.d.ts` kept reappearing locally and showing up in `git status`.
- `web/package.json` has zero Next.js deps, so nothing in the project is actively regenerating these — they're stale local cruft from before the migration (or from a tool that occasionally re-creates them).
- Adds `/.next` and `/next-env.d.ts` to `web/.gitignore` so they're permanently invisible to git.
- Deletes the local copies (neither was tracked, so no git removal needed).

## Test plan
- [ ] After merge, `git status` in a fresh clone or after a build should not list `web/.next/` or `web/next-env.d.ts`.
- [ ] `npm run build` / dev container still works (no actual Next.js consumer to break).


jeffredodd:
@claude CI needs e2e fixing please

PR #972, opened 2026-06-13
fix(scheduler): stop pushing active page when schedule disabled (#970)

### LINKED ISSUE #970: Unable to stop Page updates
**Describe the bug**
Today, my daughter drafted a custom message using the Vestaboard app and pushed it to the Vestaboard. I then toggled the schedule to "off" within the Fiestaboard software so the schedule would not overwrite her custom message (which she wanted to display when visitors arrived).

However, even though the schedule was now off, the Fiestaboard logic seems to force a specific Page to be displayed on the board (this particular Page happened to have a clock display, so it kept pushing variable updates as well).  I could not figure out how to tell the Fiestaboard software to stop pushing this page to the board. The only work around I came up with was to draft a blank Page, save it, manually set it as the active page from the Home Screen of Fiestaboard, and then, have her send the message again to the board via the Vestaboard software. Since this blank page doesn't have variables, the Fiestaboard software is no longer overwriting the board.

Ask:
We need a way to manually silence the logic of the Fiestaboard, on-demand, (either for a specific amount of time, or indefinitely until it's manually un-silenced).

Additionally, a way to control what happens when the schedule is disabled (do nothing should be possible instead of forcing a page onto the board).

**Environment**
- FiestaBoard version: [7.3.0]
- Board connection mode: [local]
- Hardware: [Raspberry Pi 3]

jeffredodd:
Closes #970

## Repro

1. Toggle the schedule off on the Schedule page (so out-of-band content sent from the Vestaboard app should be preserved).
2. Observe the polling loop continuing to push the manually-set active page (e.g. a clock page) to the board every tick, overwriting the custom message.

Pre-fix, ``DisplayService.check_and_send_active_page`` fell back to "manual mode" when ``is_schedule_enabled()`` returned False: it loaded the manually-set active page (or defaulted to the first page) and pushed it. There was no way to tell FiestaBoard "stop touching the board."

## Fix

When schedule mode is off and no explicit user override is active, the polling-loop branch now returns early. The board is left alone.

Explicit user actions are unchanged:
- ``POST /settings/active-page`` immediately pushes via its own send path
- ``POST /settings/temporary-override`` flows through ``consume_temporary_override`` above the new guard, so "show this page now" still works
- Plugin trigger overrides above the guard are also unaffected

Scope is the minimal interpretation of the bug — making "schedule off" honest. The reporter's larger ask (on-demand silence timer, post-schedule-off behavior controls) is deliberately out of scope for this fix.

## Tests

- New ``tests/test_schedule_disabled_no_push.py`` regression suite: 5 tests covering single tick, multiple ticks, no-manual-page, schedule-on-still-pushes sanity, and temporary-override-still-works.
- Updated existing silence-mode and integration tests that implicitly relied on the old manual-mode behavior to assert schedule-on (silence is a sub-feature of an active polling loop).
- Full ``pytest tests/`` run: 3721 passed, 11 skipped.

## Test plan

- [ ] Toggle the schedule off in the UI and confirm a custom message sent from the Vestaboard app stays on the board.
- [ ] Toggle the schedule back on and confirm scheduled/active pages start updating again.
- [ ] With schedule off, push a page via the active-page UI and confirm it lands on the board.
- [ ] With schedule off, push a temporary override and confirm it lands and reverts as configured.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -285,20 +285,19 @@ class DisplayService:
                         f"Schedule mode: No matching schedule for {current_day} {current_time.strftime('%H:%M')}"
                     )
             elif active_page_id is None:
-                # Manual mode: Use manual active page setting
-                active_page_id = settings_service.get_active_page_id()
-                logger.debug(f"Manual mode: Using manual active page: {active_page_id}")
-
-            # No active page set - try to default to first page (manual mode only)
-            if not active_page_id and not settings_service.is_schedule_enabled():
-                pages = page_service.list_pages()
-                if pages:
-                    active_page_id = pages[0].id
-                    settings_service.set_active_page_id(active_page_id)
-                    logger.info(f"No active page set, defaulting to first page: {active_page_id}")
-                else:
-                    logger.debug("No active page and no pages available")
-                    return False
+                # Schedule is disabled and no explicit user override is active.
+                # Issue #970: a user toggling the schedule off expects FiestaBoard
+                # to stop pushing to the board so out-of-band content (e.g. a
+                # message composed in the Vestaboard app) isn't overwritten by
+                # the polling loop. Don't auto-push the manually-configured
+                # active page here — explicit user actions (POST
+                # /settings/active-page, POST /settings/temporary-override)
+                # still flow through their own send paths, so the user retains
+                # a way to push when they want to.
+                logger.debug("Schedule disabled - skipping active-page auto-push (issue #970)")
+                self._last_active_page_content = None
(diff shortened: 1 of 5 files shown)


jeffredodd:
Closing — reopening as a feature, not a bug fix. The current PR makes `schedule.enabled = false` silently stop pushes, which overloads the schedule toggle and would surprise anyone using manual-mode-without-rotation. The right shape is a first-class per-board "Paused" toggle, surfaced in the UI. Will follow up with a new PR.

PR #919, opened 2026-06-07
feat!(web): migrate Next.js → React Router v7 (SPA) for HA Ingress

jeffredodd:
## Summary
- The Next.js frontend has been fighting Home Assistant Ingress for months (PRs #913 → #918). Each fix exposed a deeper Next.js internal: build-time `assetPrefix` has no runtime equivalent, and React 19's `ReactDOM.preload()` constructs asset URLs *after* hydration that no proxy-level rewrite or prototype patch can safely intercept without breaking React's own machinery.
- This PR replaces Next.js 16 with **React Router v7 framework mode (`ssr: false`)** — a Vite-built static SPA. The Node runtime tier is gone from production; nginx serves `/app/web/build/client/` directly. HA Ingress is now handled by a single nginx `sub_filter` that prepends `$http_x_ingress_path` to absolute `/assets/`, `/sw.js`, `/api/`, `/icons/`, `/manifest.json`, `/favicon.ico` references in HTML, JS, and CSS responses. Every URL Vite emits lives as a string literal in the build output, so the rewrite is exhaustive — no React-internal runtime URLs to chase.
- 192/200 existing regression tests pass against the new build (96%). The 2 failures are minor test-brittleness issues (timing-sensitive redirect placeholder; non-strict `getByText` matcher), not framework bugs. The 8 custom smoke tests for the new SPA pipeline all pass.

## Changes

**Framework & build**
- Next.js 16 → React Router v7 (`@react-router/dev` + `vite`). `react-router.config.ts` sets `ssr: false`; `vite.config.ts` uses `base: "/"` (absolute paths required for ES dynamic `import()` to resolve nested route chunks correctly).
- `vite-plugin-pwa` replaces `@ducanh2912/next-pwa` (Workbox `NetworkFirst` config ported 1:1).
- Tailwind v4 keeps working via `@tailwindcss/vite`.
- Self-hosted Geist fonts via `@fontsource-variable/geist` + `-mono` (HA-Ingress-safe — no external font CDN). CSS vars `--font-geist-sans` / `--font-geist-mono` moved into `globals.css`.

**Routing**
- `next/app` → 15 file-based route modules under `web/app/routes/`. Each module is a thin re-export of the existing `web/src/app/<route>/page.tsx` (single-PR mode — avoids physically moving 187 source files). Dynamic `[pluginId]` / `[id]` routes bridge `useParams()`.

**i18n**
[...]

PR #980, opened 2026-06-13
fix(pwa): immediate SW takeover, no-doc runtime cache, drop dead killer

jeffredodd:
## Summary

Companion to #979. The dev white-page symptom was the visible head of a quieter prod-side iceberg in the PWA story. Three fixes, same root cause:

- **`vite.config.ts`** — add workbox `skipWaiting: true`, `clientsClaim: true`, `cleanupOutdatedCaches: true`. A new SW shipped in a deploy currently waits for *every* controlled tab to fully close before activating, so users on long-lived FB tabs stay on the old build until they quit. After this PR, the new SW takes over immediately on update.
- **`vite.config.ts`** — narrow `runtimeCaching` to skip `request.destination === "document"`. The prior catch-all `urlPattern: /^https?.*/` + `NetworkFirst` with a 10s network timeout meant a transient slow network served a 7-day-old cached `index.html` whose hashed `/assets/<old>.js` chunks may not exist in the current build → blank screen until hard refresh. The document is already covered by precache + `navigateFallback: "/offline"` for the offline case, so excluding it from the runtime cache is the safe move.
- **`app/entry.client.tsx`** — delete `killStaleServiceWorkers` and its forced-reload boot path. It was written for the Next-PWA → vite-plugin-pwa cutover (shipped in #919 + #928, long done). Worse, the marker check (`"fiestaboard-vite-pwa"`) never matched the actual scriptURL (`/sw.js` — no marker in the URL), so **every prod user's valid current SW was being flagged stale, unregistered, caches wiped, and a `window.location.reload()` forced — once per session.** With workbox lifecycle now handled correctly, this code is both dead and harmful.

## Net effect on prod

- Faster perceived rollout — new SW activates immediately on deploy.
- One fewer forced page reload per user per session.
- Offline cache (precache + runtime cache of non-document requests) actually persists across visits.
- No more stale-HTML / missing-chunks white screen on slow networks.

## Out of scope

The `navigateFallback: "/offline"` route still serves precached `/offline` for true offline navigations. That behavior is unchanged.

## Test plan

- [ ] `npm run typecheck` and `npm run lint` clean (touched files only — pre-existing failures in `tests/note-pages.spec.ts` are unrelated)
- [ ] Build the prod image, deploy to a staging env, open in a browser that already has the current prod SW installed → after the new build is up, the new SW should activate within seconds (DevTools → Application → Service Workers should show the new version active, not waiting)
- [ ] Verify navigations during a slow-network simulation (DevTools throttling > 10s) still attempt network, never serve a stale HTML response, fall back to `/offline` only when truly offline
- [ ] First-load offline mode (after the SW is installed) → `/offline` route serves
- [ ] Boot doesn't double-reload anymore; sessionStorage no longer holds `fiestaboard:sw-killed:v1`

### DIFF
--- a/web/app/entry.client.tsx
+++ b/web/app/entry.client.tsx
@@ -2,11 +2,13 @@
  * Client entry point for the React Router v7 build.
  *
  * Runs once on first load. Responsibilities:
- *  1. Unregister any stale service workers from the previous Next-PWA
- *     deployment (kill-switch) so users don't get cached `/_next/*`
- *     responses after the cutover. This is a one-time cost per user.
- *  2. Initialize i18next via the side-effect import.
- *  3. Hydrate the React Router app.
+ *  1. Initialize i18next via the side-effect import.
+ *  2. Hydrate the React Router app.
+ *
+ * Service worker lifecycle is handled by vite-plugin-pwa's auto-injected
+ * register script (`injectRegister: "auto"` in vite.config.ts) plus the
+ * workbox `skipWaiting` / `clientsClaim` config — no manual SW management
+ * is needed here.
  */
 import "@/i18n/i18next";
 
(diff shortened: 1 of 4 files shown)


jeffredodd:
@claude Can you fix the issues with our e2e tests?

jeffredodd:
@claude — rebased onto main, which now includes #987 granting this workflow `contents: write` / `pull-requests: write`. Your previous run had the right fix but the push was denied 403. Please re-attempt the fix you proposed in [comment 4699425518](https://github.com/Fiestaboard/FiestaBoard/pull/980#issuecomment-4699425518):

- `web/tests/a11y.spec.ts` — parallelize the independent `beforeEach` ops with `Promise.all` (keep `configureBoard` → `resetToSingleBoard` sequential)
- `web/playwright.config.ts` — `retries: process.env.CI ? 1 : 0`

Push directly to this branch (`fix-prod-pwa-config`); the push should now succeed.

PR #1002, opened 2026-06-13
feat(board): adaptive post-send refresh — UI catches up in <1s, not 30s

jeffredodd:
## Summary
- **Backend** (`src/main.py`): `DisplayService.request_board_refresh()` now adaptively polls the Vestaboard after a send instead of a single shot 3s later. Initial 500ms delay, then retries every 1s, capped at 3s total. Stops early as soon as the read matches what we sent. Every successful read still updates the cache, so even a timeout improves freshness over the old behavior. A subsequent call cancels the in-flight cycle via `threading.Event` so rapid sends don't stack threads.
- **Frontend** (`web/src/hooks/use-board.ts`): on successful `useSetActivePage`, schedule `board-current-message` invalidations at 750ms and 3500ms. The early tick catches the fast local-API case; the late tick is the safety net after the backend's adaptive window closes.

## Why
Switching active pages felt slow: the backend scheduled one read 3s after a send, and the React Query that drives the board display polls every 30s. So the UI could lag 0–30s behind the physical board after every change. With these two changes the UI typically reflects the new board state within ~1s.

## Test plan
- [x] New backend tests: `tests/test_adaptive_board_refresh.py` (9 cases: early stop on match, retries until match, cache on timeout, time budget enforced, cancellation, None / exception handling, no `_last_characters`, no `vb_client`).
- [x] New frontend test in `web/src/__tests__/hooks-extended.test.ts` verifying `useSetActivePage` invalidates `board-current-message` at ~750ms and ~3500ms.
- [x] Full backend suite: 3789 passed, 11 skipped.
- [x] Full frontend suite: 1053 passed, 13 skipped.
- [x] Ruff lint + format on touched Python files.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -45,7 +45,12 @@ class DisplayService:
         self._polled_characters: list[list[int]] | None = None
         self._polled_at: float | None = None
         self._poll_thread: threading.Thread | None = None
-        self._pending_refresh_timer: threading.Timer | None = None
+        # Adaptive post-send refresh: a background thread polls the board
+        # on a short ramp after each send and stops early once the read
+        # matches what we just sent. ``_refresh_cancel`` lets a subsequent
+        # call abort an in-flight cycle so rapid sends don't pile up.
+        self._refresh_thread: threading.Thread | None = None
+        self._refresh_cancel: threading.Event | None = None
 
     def _build_board_clients(self):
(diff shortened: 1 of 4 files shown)


jeffredodd:
@claude Can you look at the CI failure and fix the listing here? Should be easy

Issue #1181, opened 2026-06-18
Note arrays: Cloud API mock for integration tests


jeffredodd:
**Part of:** #1167 · **Label:** `vestaboard-arrays`
**Blocked by:** #1168 · **Blocks:** #1179

### Summary
Add a mock of the new Cloud API (`cloud.vestaboard.com`) to `integration-tests/` so end-to-end flows can exercise note-array send/read without hitting Vestaboard.

### Context
`integration-tests/` already includes a multi-board mock (the `BoardInstance.port` field exists for "multi-board mock e2e"). Extend it with the new endpoint, the `X-Vestaboard-Token` header, and `currentMessage.layout` read responses sized to a configured array (needed by the #1172 auto-detect path).

### Scope / implementation
- Mock `POST`/`GET` for `cloud.vestaboard.com` with `X-Vestaboard-Token`:
  - `POST` stores/echoes the last `characters` grid.
  - `GET` serves `{"currentMessage": {"layout": "<json grid>", "id": ...}}` for reads.
- Allow configuring the mocked array size so auto-detect can be tested (e.g., 6×30).
- Wire the mock into the integration-test harness / compose so tests can target it.

### Acceptance criteria
- [ ] Integration tests can send a grid to, and read a layout from, a mocked note array.
- [ ] The mock validates the token header and returns a correctly sized layout.

### Tests
Exercised by #1179 and the integration suite.
```
docker-compose -f docker-compose.dev.yml exec fiestaboard pytest integration-tests/ -v
```

### Verify
Send a 6×30 grid to the mock, read it back, and confirm dimensions + token enforcement.

Issue #1172, opened 2026-06-18
Note arrays: auto-detect board type & size (flagship/note/array)


jeffredodd:
**Part of:** #1167 · **Label:** `vestaboard-arrays`
**Blocked by:** #1168 · **Blocks:** #1178

### Summary
Add a backend endpoint that reads a board's current layout via a `GET` and **infers both its device type and its size** — `flagship` (6×22), `note` (3×15), or a note array (`notes_wide × notes_tall`). Motivated by note arrays, but it works for **any** board, because every board returns a layout grid we can measure.

### Context
A `GET` of the current message returns the layout grid for any board:
- Local API / RW Cloud API (`rw.vestaboard.com`) for `flagship`/`note` — the existing read path (`parse_read_message_payload`, `src/board_client.py` lines 92-117).
- The new Cloud API (`cloud.vestaboard.com`) for note arrays — read path from #1168.

From the grid's `rows × cols` we classify the board: `6×22 → flagship`, `3×15 → note`, otherwise if `rows%3==0 && cols%15==0` → note array with `notes_wide = cols/15`, `notes_tall = rows/3`.

### Scope / implementation
- Add a shared classifier in `src/devices.py`: `classify_dimensions(rows, cols) -> {device_type, notes_wide?, notes_tall?, matched_preset?}` (reuses `is_valid_note_array_grid` and the preset table from #1183).
- Add a route in `src/api_server.py`, e.g. `POST /api/boards/{board_id}/detect-size`, that:
  - reads the current layout using the **board's own api_mode** (local / RW cloud / note-array cloud),
  - runs `classify_dimensions`, and
  - returns `{ device_type, notes_wide?, notes_tall?, rows, cols, matched_preset? }`.
- Handle error cases with clear, structured messages: board unreadable / not configured, empty board (no current layout), and dimensions that don't classify cleanly.

### Out of scope (other issues)
The Auto-detect **button** + wiring → #1178 · the note-array read transport → #1168.

### Acceptance criteria
- [ ] `6×22` layout → `flagship`; `3×15` → `note`.
- [ ] `6×30` → note array `2 × 2`, preset "2×2 grid"; `3×60` → `4 × 1`, preset "4 side-by-side".
- [ ] Works against a flagship/note board (existing read path) and a note-array board (#1168 read path).
- [ ] Unreadable / empty / unclassifiable boards return a clear, structured error (not a 500).

### Tests
[...]

Issue #1167, opened 2026-06-18
Epic: Support Vestaboard Note Arrays end-to-end


jeffredodd:
## Epic: Support Vestaboard Note Arrays end-to-end

Vestaboard shipped **Note Arrays** — 2+ Vestaboard Notes ganged into a single logical device with non-standard dimensions, driven by a new Cloud API. This epic tracks adding end-to-end support to FiestaBoard: model the sizes, send/read via the new Cloud API, render correctly at any size, and give the web UI a variable, scrollable board preview plus configuration (presets, custom, auto-detect).

> Docs: [note-arrays](https://docs.vestaboard.com/docs/read-write-api/note-arrays) · [endpoints](https://docs.vestaboard.com/docs/read-write-api/endpoints/) · [product page](https://www.vestaboard.com/note-array)

### What is a note array?
A note array is an **N notes-wide × M notes-tall** grid. Each Note is **15 cols × 3 rows**, so `rows = notes_tall × 3` and `cols = notes_wide × 15`. The Cloud API accepts arbitrary nested character arrays ("installations of any size or shape"), so we model the **general case** and ship the 5 documented arrangements as **presets** plus a **Custom** option.

| Preset | notes W×H | Dims (rows×cols) |
|---|---|---|
| 2 side-by-side | 2 × 1 | 3 × 30 |
| 4 side-by-side | 4 × 1 | 3 × 60 |
| 2 stacked | 1 × 2 | 6 × 15 |
| 4 stacked | 1 × 4 | 12 × 15 |
| 2×2 grid | 2 × 2 | 6 × 30 |

### API & constraints
- **Send:** `POST https://cloud.vestaboard.com/`, header `X-Vestaboard-Token`, body `{"characters": [[...]]}` — distinct from the Read/Write API (`rw.vestaboard.com`) used today.
- **Read:** `GET https://cloud.vestaboard.com/` → `currentMessage.layout` (JSON string), sized to the array.
- **No transitions** are supported on note arrays.
- **Rate limit:** ≤ 1 message / 15s (more may be dropped).

### Modeling decision
Generalized model: store `notes_wide × notes_tall` on the board; derive dimensions. Reuse the existing `DEVICE_DIMENSIONS` / `device_type` machinery in `src/devices.py` (already threaded through page validation, the template engine, grid validation, and the web preview). Presets are shortcuts to W×H pairs; Custom sets W×H directly; auto-detect infers W×H from the board's current layout.

### v1 scope
[...]

PR #1223, opened 2026-06-19
feat(board): note-array send constraints — no transitions, 15s throttle [#1169]

### LINKED ISSUE #1169: Note arrays: send constraints (no transitions, 15s rate limit)
**Part of:** #1167 · **Label:** `vestaboard-arrays`
**Blocked by:** #1168 · **Blocks:** #1179

### Summary
Enforce the two hard constraints the Vestaboard docs place on note arrays: **transitions are not supported**, and **no more than one message every 15 seconds** (extra messages may be dropped).

### Context
`send_characters` in `src/board_client.py` (lines 232-324) supports transition strategies (`VALID_STRATEGIES`, lines 54-64). For note-array boards these must be omitted. The cloud also drops messages sent faster than 1 / 15s, so we need a client-side guard so schedule/polling loops don't silently lose updates.

### Scope / implementation
- In the send path, when the board `is_note_array`, never include transition params (ignore/strip any requested strategy; log at debug).
- Add a **per-board minimum send interval** (≥ 15s) for note arrays: track the last-send timestamp keyed by board id; if a send arrives sooner, skip it (recommended) and log a warning, returning a clear status to the caller. Document the chosen behavior (skip vs. delay) in the PR.
- The schedule/polling loops in `src/api_server.py` call the client; the client-level guard is the safety net, but note any loop that could exceed the rate so reviewers can confirm.

### Out of scope (other issues)
Client endpoint itself → #1168.

### Acceptance criteria
- [ ] Note-array sends never include a transition strategy.
- [ ] A second send within 15s of the previous one is throttled (logged, not sent), verified with a mocked/injected clock.
- [ ] Flagship/note behavior (including transitions) is unchanged.

### Tests
Extend `tests/test_board_client.py`: assert no-transition for note arrays; assert throttle with a controllable clock.
```
docker-compose -f docker-compose.dev.yml exec fiestaboard pytest tests/test_board_client.py -v
```

### Verify
Two rapid note-array sends → only the first reaches the (mocked) transport; flagship sends still carry transitions.

jeffredodd:
Closes #1169. Part of the Note Arrays epic #1167. Targets `next`.

## What
Enforces the two Cloud-API constraints for note-array boards in `BoardClient.send_characters`:

- **No transitions** — note-array sends strip any `strategy` / `step_interval_ms` / `step_size` (forced to `None`) and log a debug note; the payload + transition-info log can never re-introduce them.
- **≥15s rate limit** — a module-level `_note_array_last_send` dict keyed by the board's `note_array_token` throttles re-sends within `NOTE_ARRAY_MIN_SEND_INTERVAL` (15s), returning `(True, False)` and warning. The dict is module-scoped so it survives `BoardClient` recreation (clients are rebuilt on reinit / per-call), matching the server-side, per-board nature of the limit. Time is injectable (`_time_func`, defaults to `time.monotonic`) for deterministic tests.

## Tests
`TestNoteArrayConstraints` in `tests/test_board_client.py`: transition-strip, throttle within/at/after 15s, first-send pass-through, persistence across client recreation, and a flagship-still-has-transitions regression. An autouse fixture resets the throttle dict between tests.

## Verification
- `ruff check` / `format`: clean
- targeted `tests/test_board_client.py`: **118 passed**
- broader suite (minus `tests/ai`): **3817 passed, 0 failures**

Disjoint from the other in-flight Wave-3 PRs (#1172, #1181).

### DIFF
--- a/src/board_client.py
+++ b/src/board_client.py
@@ -16,6 +16,8 @@ Cloud API Reference:
 import json
 import logging
 import re
+import time as _time_module
+from collections.abc import Callable
 from typing import Any, Literal, Optional
 
 import requests
@@ -62,6 +64,13 @@ TransitionStrategy = Literal[
 ]
 
 VALID_STRATEGIES = ["column", "reverse-column", "edges-to-center", "row", "diagonal", "random"]
+
+# Minimum interval (seconds) between note-array sends enforced client-side.
+NOTE_ARRAY_MIN_SEND_INTERVAL: float = 15.0
+
+# Module-level per-board throttle state. Key = note_array_token (board id proxy).
+# Persists across BoardClient recreations within a process.
+_note_array_last_send: dict[str, float] = {}
 
 
 def _valid_grid_dimensions() -> set:
@@ -150,6 +159,7 @@ class BoardClient:
         note_array_token: str | None = None,
         notes_wide: int = 1,
         notes_tall: int = 1,
+        _time_func: Callable[[], float] | None = None,
     ):
         """
         Initialize board API client.
@@ -163,6 +173,8 @@ class BoardClient:
             note_array_token: X-Vestaboard-Token for note-array boards; non-empty enables note-array mode.
             notes_wide: Number of Notes side-by-side (columns = notes_wide * 15).
             notes_tall: Number of Notes stacked (rows = notes_tall * 3).
+            _time_func: Injectable monotonic clock for note-array throttle tests.
+                Defaults to ``time.monotonic``.
         """
         if not api_key:
             raise ValueError("api_key is required")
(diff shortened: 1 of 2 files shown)


jeffredodd:
@claude please review for correctness and edge cases. Focus: (1) is a module-level throttle dict keyed by `note_array_token` the right scope given clients are recreated on reinit, vs an instance field; (2) the strategy-strip belt-and-suspenders — any path that could still leak a transition into the note-array payload; (3) the throttle return value `(True, False)` semantics (treated as success-but-unchanged) vs returning a distinct 'throttled' signal. Please **post findings as review comments only — do not push commits**; I'll apply revisions as the maintainer.

PR #1254, opened 2026-06-20
Disney Parks Queue Times Control Update

DarthXoc:
## Description
For the _Disney Parks Queue Times_ plugin, I've added the ability for users to set custom names for rides. Additionally, users can now reorder rides without having to delete all of their current rides and adding the rides again in display order.

Both of these changes are gated so that they won't appear when configuring the plugin unless the user has an updated version of the plugin installed that contains the prerequisite feature flags (e.g. `customRideNames` and `reorderRides`).

These changes had to be made inside the core project as the _Disney Parks Queue Times_ plugin does not contain the actual controls that render to configure the integration.

A separate PR for the _Disney Parks Queue Times_ plugin is incoming, which will require FiestaBoard `7.10.0` (or higher).

## ⚠️ FiestaBoard Version
- FiestaBoard version must be set to 7.10.0 (or higher).

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] New plugin
- [ ] Documentation update
- [ ] Refactor / maintenance

## Checklist
- [x] I have tested my changes locally
- [x] I have added/updated tests where applicable
- [x] I have updated documentation where applicable
- [x] My code follows the project's coding standards


jeffredodd (approved):
Thanks for the contribution, Jason, this is really clean work. The translations across all 14 locales, the interpolated aria-labels on the new controls, and gating the custom-name and reorder features behind capability flags so older plugin versions degrade gracefully are all exactly what we look for. The new tests cover the behavior nicely too.

I pushed one small commit to run Prettier on the three changed files so the lint-web check goes green. It was just JSX indentation and line wrapping, no logic changes.

Approving and enabling auto-merge. Thanks again!

Issue #1248, opened 2026-06-20
Per-board #7: unify Schedule page with shared current board


jeffredodd:
Part of #1241. Depends on #5.

## Unify the Schedule page with the shared current board

The Schedule page already has a local, non-persisted board selector (`selectedBoardId` state that resets on reload). Replace it with the shared `useCurrentBoard()` so the Schedule and Dashboard stay in sync and the selection persists.

### Requirements

1. In `web/app/routes/schedule.tsx`, replace the local `selectedBoardId`/`effectiveBoardId` state and its `useEffect` defaulting with `useCurrentBoard()`.
2. Keep the board-scoped queries (`["schedules", boardId]`, `api.getSchedules(boardId)`, etc.) driven by the shared `currentBoardId`.
3. Remove the now-redundant in-page board `<Select>` (the sidebar selector replaces it) — or keep a compact read-only board indicator. Ensure single-board installs are unaffected (no selector, default board).

### Key files / reuse

`web/app/routes/schedule.tsx`, `useCurrentBoard()` (#5), existing `queryKeys.schedules(boardId)` in `web/src/hooks/use-board.ts`.

### Acceptance criteria

- Switching the sidebar board re-scopes the Schedule page; selection persists across reloads and matches the Dashboard.
- No duplicate/competing board selector remains.

### Tests

Playwright: switch board in the sidebar → Schedule shows that board's schedules; persists across reload. `vitest` where applicable. Run web tests in Docker.

### Dependencies

#5 (`useCurrentBoard`).

jeffredodd:
Done — landed in PR #1260 on `next`, now on `main` via the #1185 squash merge (f713a899). Verified on main: `web/app/routes/schedule.tsx` consumes the shared `useCurrentBoard()`; local selectedBoardId state removed.

Issue #1246, opened 2026-06-20
Per-board #5: sidebar board selector + CurrentBoardProvider


jeffredodd:
Part of #1241. Depends on #1.

## Sidebar board selector + `CurrentBoardProvider`

Add a persisted "currently managed board" selection and a sidebar selector that works in collapsed mode. This is the frontend foundation consumed by the Dashboard (#6) and Schedule (#7).

### Requirements

1. **`CurrentBoardProvider`** (new): React context holding `currentBoardId`, persisted to `localStorage` (key e.g. `fiestaboard_current_board`), defaulting to the **primary board** (`boards[0].id`), reconciled against the live board list (drop a stale id). Expose `useCurrentBoard()` → `{ currentBoardId, setCurrentBoardId, currentBoard, boards }`. Mount in `web/src/components/providers.tsx`. Mirror the `SidebarProvider`/`sidebar-context.tsx` pattern.
2. **Sidebar selector** in `web/src/components/navigation-sidebar.tsx`: render only when `boards.length > 1`. Use the existing `ui/select` (`SelectTrigger/SelectValue/SelectContent/SelectItem`). **Collapsed mode:** icon-only trigger + `ui/tooltip` showing the board name (follow the existing collapsed nav pattern: `opacity-0 max-w-0` for the label). Place between primary nav and the secondary/footer block. Works in the mobile drawer too.
3. **i18n + a11y:** add keys (e.g. `navigation.boardSelector`, `navigation.selectBoard`, a board-name fallback) to `web/messages/en.json` **and all 14 locales** (de, es, fr, it, ja, ko, nl, pl, pt, ru, sv, tr, zh) — CI enforces parity. `aria-label` on the trigger; follow existing nav a11y conventions.

### Key files / reuse

`web/src/components/navigation-sidebar.tsx`, `web/src/components/sidebar-context.tsx` (pattern), `web/src/components/providers.tsx`, `web/src/components/ui/select.tsx`, `web/src/components/ui/tooltip.tsx`, `web/src/hooks/use-board.ts` (`useBoardSettings`), `web/messages/*.json`.

### Acceptance criteria

- Selector appears only with >1 board; hidden for single-board installs.
- Selection persists across reloads and is shared app-wide via `useCurrentBoard()`.
- Renders correctly collapsed (icon + tooltip) and in the mobile menu; i18n parity + a11y pass.

### Tests

[...]

Issue #1241, opened 2026-06-20
Epic: Per-board scheduling, driving & sidebar board selector


jeffredodd:
### Why

PR #1185 ("Note Arrays — multi-board W×H support") added multi-board support, but it was only ever wired into **Settings → Hardware**. Everywhere else still assumes a single implicit "primary board":

- The display/driver (`src/main.py` `DisplayService`) holds **one** `self.vb_client`, built only from `boards[0]` — and `_build_board_clients` even misses note-array boards (it checks `local_api_key or cloud_key`, not `note_array_token`).
- The **manual active page** (`settings.active_page.page_id`) is global.
- The **Schedule** page's board selector is local component state (resets on reload, not shared with the Dashboard); the Dashboard has no selector.
- There is **no page↔board size compatibility** — any page can be scheduled on any board, silently truncating on the wall.

We are confident there is **no multi-board consumer in production yet**, so this is the moment to make the model correct before there is. Breaking changes are acceptable; single-board installs must migrate with **zero behavior change**.

### Target model

- **Per-board:** display engine, active page, schedule + `schedule_enabled`, send routing, MQTT/trigger delivery, pause, board-state polling.
- **Global (instance library):** Pages and Collections (one shared library; never duplicated per board).
- **Current board:** a persisted **sidebar selector** sets the "currently managed board" (frontend context + `localStorage`, default primary, shown only when `boards.length > 1`). It scopes the Dashboard + Schedule. Pages/Collections lists stay global; their pickers filter by the current board's size.

### Reuse (already exists)

`ScheduleEntry.board_id`, `schedule_service.get_active_page_id(board_id=…)`, `schedules.json.default_page_by_board`, per-board `paused`/`schedule_enabled`, `board_client_from_board_dict()`, `resolve_dimensions`/`classify_dimensions`, `web/src/lib/board-dimensions.ts`, `ui/select` + `ui/tooltip`.

### Children (dependency order)

Backend issues share `main.py`/`settings/service.py`/`api_server.py` and land **serially**; frontend runs in parallel where deps allow.

- [ ] #1242 — Foundation: settings `schema_version` + per-board active page + primary-board helper + `schedule_enabled` authority
- [ ] #1243 — Per-board display engine (`DisplayService` → `runtimes`) — depends on #1242
- [ ] #1244 — Per-board send routing (endpoints + MQTT + triggers) — depends on #1243
[...]

PR #1260, opened 2026-06-21
feat(web): unify Schedule page with shared current board [#1248]

jeffredodd:
Part of the per-board scheduling epic #1241 (issue #1248).

Replaces the Schedule page's local, non-persisted `selectedBoardId` with the shared `useCurrentBoard()` (CurrentBoardProvider from #1246) so the Schedule and Dashboard stay in sync and selection persists across reloads.

- Board-scoped queries/mutations now driven by `currentBoardId`.
- Removed the redundant in-page board `<Select>`; added a compact read-only active-board indicator (`data-testid="active-board-indicator"`, multi-board only).
- Single-board installs unchanged (`effectiveBoardId` stays undefined → unscoped default endpoints, no indicator).
- Updated `multi-board-schedule.spec.ts` to drive board switching via the shared sidebar selector.

Verified: full vitest 1166 passed / 0 failed; eslint 0 errors; prettier clean. Playwright validated by CI.

### DIFF
--- a/web/app/routes/schedule.tsx
+++ b/web/app/routes/schedule.tsx
@@ -6,12 +6,14 @@ import {
   CalendarDays,
   List,
   MapPin,
+  Monitor,
   Plus,
   Power,
 } from "lucide-react";
 import { lazy, Suspense, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { toast } from "sonner";
 
+import { useCurrentBoard } from "@/components/current-board-context";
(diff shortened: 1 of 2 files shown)


jeffredodd:
@claude Please review for correctness and consistency with the per-board epic (#1241). Post findings as comments only — do NOT push commits. Focus: useCurrentBoard wiring, query/mutation re-scoping on board switch, and single-board no-op behavior.

PR #1288, opened 2026-06-26
feat(collections): add random selection mode (#1287)

### LINKED ISSUE #1287: Collection Page Randomizer
No

Add the ability for a Collection to randomly select any one of its assigned pages for viewing, with a new selection being made when Page Duration has expired. This could be done with a RNG running in the background

Something like this, but in Python, not Swift:
```
let pageIndex: Int = .random(in: 0..<3) // With '3' being the total number of pages
self.setPage(index: pageIndex)
```

I looked briefly at the 'Variable / expression' mode, but that seems overly complicated and this would allow a general purpose random page picker.

**Additional context**

jeffredodd:
## Summary

Closes #1287. Adds a third Collection **selection mode — `random`** — that picks a page at random each time the page duration expires, instead of cycling in order.

The implementation deliberately mirrors the existing **`time`** mode: it's **stateless and deterministic**, seeded purely by a time-derived "round" number. So it survives restarts with zero stored state and needs no background RNG loop — it just computes the right page from the clock, exactly like time mode does today.

### Behavior (decided up front)
- **No back-to-back repeats.** It's a *shuffle-bag*: each round is a seeded permutation of all pages, so every page is shown once before any repeats, and the round boundary is patched so a page is never shown twice in a row. (With exactly two pages, "no repeats" necessarily means strict alternation, which is handled directly.)
- **Page Duration** reuses the same control/semantics as time mode (`random.interval_seconds`).

Demonstrated end-to-end (4 pages, 30s windows):
```
stocks → clock → weather → news → clock → news → weather → stocks → ...
back-to-back repeats: NONE   |   all pages used: yes   |   restart-safe: identical sequence from a fresh instance
```

## Changes

**Backend** (`src/collections/`)
- `models.py`: `SelectionMode` gains `"random"`; new `RandomModeConfig`; `Collection.random` field + validator; `current_page_index_random` / `current_page_id_random` (shuffle-bag).
- `service.py`: `resolve_page_id` + `seconds_until_next_check` random branches; `create_collection` plumbs the config.
- No schema migration — the new field is optional and backward compatible.

**Web** (`web/`)
- `api.ts` / `ai-chat-types.ts`: `"random"` mode + `RandomModeConfig` type.
- `collections.tsx`: **Random** option in the selection-mode dropdown (Shuffle icon), page-duration control, and list display.
- i18n keys added across **all 14 locales**.

## Testing (TDD)
- **Python:** 12 new tests in `tests/test_collections.py` — no back-to-back repeats (n = 2,3,4,5,7), window stability, restart-safety/determinism, single-page edge case, full-coverage, and service-level resolution + timing. Written test-first (watched them fail on missing `RandomModeConfig`, then implemented). Full collections suite: **92 passed**; full platform suite green (the only failures were harness artifacts from an isolated test sandbox, confirmed passing in the full container).
- **Web:** new random-mode type-shape test; collection-related vitest specs **88 passed**; `typecheck` clean on changed files; `prettier --check` clean; `eslint` 0 errors.

## Out of scope / possible follow-up
- The MCP `create_collection` / `update_collection` tools still expose only `time`/`variable`. Not a regression; could add `random` in a follow-up.


jeffredodd:
@claude can you do a review and make any fixes you think are needed?

PR #1185, opened 2026-06-18
feat!: Note Arrays — multi-board (W×H) support [Epic #1167]

jeffredodd:
# Note Arrays — multi-board (W×H) support · Epic #1167

Adds first-class support for **Vestaboard Note Arrays**: grids of multiple Note boards driven as one logical display (2 side-by-side, 4 stacked, a 2×2 grid, or a custom layout up to 8×8 Notes). Previously FiestaBoard only knew two fixed sizes — Flagship (22×6) and Note (15×3). The whole stack — model, Cloud API transport, rendering, settings, preview, auto-detect, plugins, tests, docs, and a local dev mock — is now **dimension-driven**.

This branch also **reconciles with `main`**, including @DarthXoc's **Device Size Support (#1232)** (plugins read a board's width/height), and **extends it with note-array awareness**.

## Issues closed on merge

Closes #1167 (epic) and every sub-issue:

- Closes #1168 — Cloud API send/read in BoardClient
- Closes #1169 — send constraints (no transitions, 15s rate limit)
- Closes #1170 — credentials & board config
- Closes #1171 — de-hardcode API-server geometry
- Closes #1172 — auto-detect board type & size
- Closes #1173 — render correctly at any dimensions
- Closes #1174 — TypeScript types + shared dimension helpers
- Closes #1175 — board-size indicator UI
- Closes #1176 — variable-size board preview
- Closes #1177 — hybrid fit-to-width / actual-size preview with scroll
- Closes #1178 — settings: presets, custom W×H, token, auto-detect
- Closes #1179 — Python test suite
- Closes #1180 — web regression (Playwright)
- Closes #1181 — Cloud API mock for integration tests
- Closes #1182 — documentation
- Closes #1183 — dimension model (generalized W×H)

---

## What's included

**Model & geometry**
- Generalized W×H dimension model: a Note = 15 cols × 3 rows; an array = `notes_wide×15` cols × `notes_tall×3` rows; capped at 8 Notes per axis (#1183).
- 5 presets — 2 side-by-side (3×30), 4 side-by-side (3×60), 2 stacked (6×15), 4 stacked (12×15), 2×2 grid (6×30) — plus Custom W×H (#1183).
- De-hardcoded the API server / formatter geometry so `/fill`, `/info`, welcome, etc. size to the active board (#1171).
- Pages render correctly at any dimension, with a `pages.json` schema migration for the new W×H fields (#1173).

**Cloud API transport**
[...]

Issue #1247, opened 2026-06-20
Per-board #6: dashboard per-board active display


jeffredodd:
Part of #1241. Depends on #5, #3.

## Dashboard per-board active display

Make the home "Active Display" reflect the **current board** (from `useCurrentBoard()`), showing that board's live display and its own active page.

### Requirements

1. Consume `useCurrentBoard()` in `web/src/components/active-page-display.tsx` / `web/app/routes/home.tsx`.
2. Fetch the **selected board's** current display + active page (board-scoped API calls using the `board_id` params added in #3). Board-scope the relevant TanStack query keys in `web/src/hooks/use-board.ts` (e.g. `activePage(boardId)`, current-message by board).
3. "Change Page" sets the **current board's** active page (board-scoped `set_active_page`), respecting the size-compat rules from #4.
4. Indicate which board is shown; keep the existing paused-boards badge.

### Key files / reuse

`web/src/components/active-page-display.tsx`, `web/app/routes/home.tsx`, `web/src/hooks/use-board.ts`, `web/src/lib/api.ts` (board_id params), `useCurrentBoard()` (#5).

### Acceptance criteria

- Switching the sidebar board updates the Active Display to that board's current tiles + active page.
- "Change Page" affects only the selected board.
- Single-board installs behave as before.

### Tests

`vitest` for board-scoped fetching/selection; Playwright: switch board → dashboard reflects it; change page targets the selected board. Run web tests in Docker.

### Dependencies

#5 (`useCurrentBoard`), #3 (board-scoped active-page + current-message endpoints).

Issue #1244, opened 2026-06-20
Per-board #3: per-board send routing (endpoints + MQTT + triggers)


jeffredodd:
Part of #1241. Depends on #2.

## Per-board send routing (endpoints + MQTT + triggers)

Route every send to a specific board via the `get_board_client(board_id)` seam added in #2, instead of the single `service.vb_client`.

### Requirements

1. **API send/refresh endpoints** accept an optional `board_id` (default → primary) and route to that board's client at that board's size:
   - `POST /api/pages/{page_id}/send`, `POST /api/send`, `POST /api/refresh` (`src/api_server.py`).
2. **Active-page endpoints** accept `board_id`: `GET/PUT /settings/active-page` (forward to the per-board getter/setter from #1).
3. **MQTT handlers** (`src/mqtt/commands.py`, `src/mqtt/state.py`): `send_message`, `blank_board`, `refresh_display`, `active_page` accept/resolve a board (payload `board_id` or board name → id; default primary), size the grid to that board, and send via its client. Replace the hardcoded 6×22 blank grid with the board's actual dims.
4. **Per-board status surfacing:** the status endpoint(s) report per-board `{configured, paused, active_page_id}` using the engine's `get_runtime`/`board_status` helpers.
5. Preserve back-compat: omitting `board_id` everywhere behaves exactly as today (targets the primary board).

### Key files / reuse

`src/api_server.py` (`_get_board_client` ~L6019, send/refresh/active-page routes), `src/mqtt/commands.py`, `src/mqtt/state.py`; `get_board_client`/`get_runtime` from #2; `resolve_dimensions` for sizing.

### Acceptance criteria

- Sending a page to board A vs board B lands on the correct mock at the correct size.
- MQTT send/blank/active-page target the specified board; default is primary.
- All existing single-board API/MQTT tests pass unchanged (default routing).

### Tests

`pytest` cases for board-scoped send/refresh/active-page routing and MQTT per-board handling (default-primary + explicit board_id), against the stdlib mocks. Run in Docker.

### Dependencies

#2 (`get_board_client` seam, runtimes).

Issue #1243, opened 2026-06-20
Per-board #2: per-board display engine (DisplayService → runtimes)


jeffredodd:
Part of #1241. Depends on #1.

## Per-board display engine

Refactor `src/main.py` `DisplayService` to independently drive **every** enabled+configured board, where today it drives a single `self.vb_client` built from `boards[0]`.

### Requirements

1. **Per-board runtimes.** Introduce a `BoardRuntime` (client + per-board cached state: `last_active_page_content/_id`, `last_silence_mode_active`, `snoozing_message_sent`, `polled_characters/_at`, refresh cancel event, read/collection cadence gates). `self.runtimes: dict[board_id, BoardRuntime]`; `self._primary_board_id`.
2. **Back-compat shims.** Keep `self.vb_client` and `self._polled_characters/_at` as **properties** aliasing the primary runtime so untouched callers (`api_server.py`, `mqtt/commands.py`) keep working. `self.vb_client = X` (used by tests) must create/replace the primary runtime. `check_and_send_active_page()` becomes a thin wrapper calling `check_and_send_for_board(primary)`.
3. **Build all boards.** Rewrite `_build_board_clients()` to populate `runtimes` for every enabled board via `board_client_from_board_dict()`, gated by **`BoardInstance.is_connection_configured`** (fixes the current note-array miss). Preserve the legacy `Config`-only single-board path.
4. **Extract `check_and_send_for_board(board, rt)`** from the current monolith (`main.py:231-465`): every `self._last_*`→`rt.*`, `self.vb_client`→`rt.client`, `board_id = board["id"]` (not `_get_first_board_id()`). Use the per-board active-page seam from #1 (`get_active_page_id(board_id)`).
5. **`_tick()`** iterates enabled boards with **partial-failure isolation** (one board raising never blocks others) and registers with `schedule` in `run()` (replacing the single job). The 1-second silence-boundary detector and collection-cadence paths call `_tick()`.
6. **Decisions:** `temporary_override` evaluated for the **primary board only** (global consume-once store); **triggers** evaluated globally once per tick, routed to each non-paused/non-silenced board sized per board; **silence** is a global decision with **per-board delivery** sized via the board's geometry.
7. **Polling:** keep **one** tick loop + **one** board-read thread iterating runtimes with per-board interval gates (don't spawn a thread per board — preserves the single-threaded send invariant the note-array ≥15s throttle relies on). Per-board `request_board_refresh`.
[...]

PR #1296, opened 2026-06-27
fix(note-arrays): size send paths to real dimensions + harden tests [gate for #1185]

jeffredodd:
## Why this PR

This is the **human-in-the-loop gate** for #1185 (Note Arrays). It targets `next`, so merging it updates #1185. Goal: when this merges, #1185 is actually **100% correct for arrays + multi-board**, with tests you can trust.

I rebuilt the `next` (`04c4f08b`) dev stack from scratch (with the new mock Cloud board on `:19200`) and verified the feature at runtime — not just by reading tests. That surfaced a **ship-blocker** the existing tests hid.

## 🔴 Blocker fixed: note-array boards 500 on every send path

`get_dimensions()` only knows `flagship`/`note` and **raises `ValueError` on `note_array`**. Five send paths called it. Reproduced live:

```
POST /templates/render/live  (note-array board)  ->  HTTP 500
  File "src/api_server.py", line 7419, in render_template_live
    dims = get_dimensions(device_type)
  ValueError: Unknown device type: note_array. ... use resolve_dimensions().
```

A user **cannot send to a note-array board** via the live editor, `/pages/{id}/send`, `/settings/active-page`, or `/displays/{type}/send`. The send-endpoint tests `patch("…get_dimensions")` and hardcode `rows=6, cols=22`, so they could never catch it.

**Fix:** all five send sites use `resolve_dimensions(device_type, notes_wide, notes_tall)`. After the fix the same live send returns **200** and lands a **3×60** grid on the mock.

## What's in this PR (all verified green)

| Change | File |
|---|---|
| 5 send paths size to real note-array dims (was 500) | `src/api_server.py` |
| Guard `from_dict` against a JSON-`null` `note_array_token` (`None.strip()` crash) | `src/devices.py` |
| **New** regression test: live send to note-array board, resolver NOT mocked, asserts 3×60 grid | `tests/test_api_extended.py` |
| Repoint 7 send tests `get_dimensions`→`resolve_dimensions` | `tests/test_api_coverage.py` |
| Make 2 note-array Cloud-URL tests **hermetic** (pass in `docker exec … pytest`, not only CI) | `tests/test_board_client.py` |
| Run the `integration-tests/` mock-cloud contract suite in CI (was never collected) | `.github/workflows/ci.yml` |
| **i18n:** localize size-indicator preset labels (English leaked into visible text **and** aria-label in 13 locales) | `web/src/components/board-size-indicator.tsx` |
| **Type contract:** add `schedule_enabled` + `port` to `BoardInstance` (server emits them; TS type was incomplete) | `web/src/lib/api.ts` |

**Results:** Python `4102 passed / 11 skipped` (was `4099 + 2 failing`), integration `25 passed`, web vitest `1166 passed`, prettier/eslint clean.

## Verification performed against the live `next` build

- **Cloud transport (14/14, real BoardClient → real mock):** URL + `X-Vestaboard-Token`, body exactly `{"characters": grid}` with transitions stripped, 3×60 lands, `read` round-trips, **15s throttle** blocks then allows, **per-board token isolation** holds.
- **Auto-detect (13/13):** every preset + custom 8×8 + flagship + note classify correctly; 404/422 paths; token masked.
- **Web E2E (78 passed):** note-arrays regression, note-pages, multi-board, multi-board-schedule (incl. `board_id` filtering + board-switch isolation), setup wizard, device-mismatch.
- **Pages schema migration** v3→v4: correct, idempotent, no data loss.
- **TS↔Python contract** (full sweep): in sync except the two `BoardInstance` fields fixed here.
- **i18n parity:** all 14 locales carry every note-array key (no missing/orphan); the one real leak is fixed here.

## ✅ Investigated and cleared (were flagged, verified NOT bugs)

- **Settings migration "drops `schedule.enabled`":** not a bug. `is_schedule_enabled()` falls back to the global mirror when there are **no** boards, and the migration stamps per-board `schedule_enabled` when boards exist (and never deletes the global value). No upgrade data loss.
- **"Re-POSTing boards drops `schedule_enabled`/`port`":** not data loss. The Settings UI spreads the received board object (`{...b, ...updates}`), so runtime fields survive even though the TS type omitted them (now added for honesty).

## ⚠️ Still your call (out of scope / not arrays-correctness)

1. **Multi-board driving is primary-only (confirm intended).** The polling loop drives only the first board (`src/main.py:733`) and the dashboard preview ignores the sidebar selection (`active-page-display.tsx`); also `PUT /settings/active-page` is primary-board-scoped (no `board_id`). Looks like later per-board-epic scope, not claimed-broken here.
2. **Unpinned ruff (`ruff>=0.9.0`) → nondeterministic CI format-check.** ruff 0.15 reformats **64 src/ files** of pre-existing code. Can fail `ruff format --check` for any new PR to `next`. Recommend pinning ruff + a one-time reformat (separate PR).
3. **detect-size "unclassifiable" 422 branch is dead** — the read layer pre-filters odd grids to `None`, so users get a misleading "blank or unreachable" message for an unsupported size.
4. **Lower-priority test-quality gaps** (finders, not all verified): `scaled-board-display` fit-mode scaling untested; seam *gap* not asserted; `page-grid-selector` renders note-array thumbnails at 1×1; plugin note-array cache tests assert only `fetch_count`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,7 +194,17 @@ jobs:
         pytest -m integration tests/test_plugin_install_integration.py -v
       env:
         BOARD_READ_WRITE_KEY: test_key
-    
+
+    - name: Run note-array Cloud mock contract tests
+      run: |
+        # The mock-cloud Cloud API contract suite lives under integration-tests/,
+        # which is excluded from [tool.pytest.ini_options] testpaths, so the main
+        # test step above never collects it. It is stdlib-only and self-hosts its
+        # mock HTTP servers (no Docker needed), so it can run directly on the host.
+        pytest integration-tests/ -v
+      env:
+        BOARD_READ_WRITE_KEY: test_key
+
     - name: Upload platform coverage
       uses: codecov/codecov-action@v7
       with:
--- a/src/api_server.py
+++ b/src/api_server.py
@@ -36,7 +36,7 @@ from .collections.models import CollectionCreate, CollectionUpdate, is_collectio
 from .collections.service import get_collection_service  # noqa: E402
 from .config import Config  # noqa: E402
 from .config_manager import get_config_manager  # noqa: E402
-from .devices import classify_dimensions, get_dimensions, resolve_dimensions  # noqa: E402
+from .devices import classify_dimensions, resolve_dimensions  # noqa: E402
 from .displays.service import get_display_service, reset_display_service  # noqa: E402
 from .main import DisplayService  # noqa: E402
 from .network.wifi import WiFiError, get_wifi_service  # noqa: E402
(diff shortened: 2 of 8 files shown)


jeffredodd:
**Multi-board scope clarification** (resolves flagged item #1): the per-board *driving* limitations I noted are tracked as separate **open** epic issues, not gaps in #1185's claims:

- Polling loop drives primary board only (`src/main.py:733`) → **#1243** (per-board display engine) + **#1244** (per-board send routing)
- Dashboard preview / manual active-page ignore the sidebar-selected board (`active-page-display.tsx`; `PUT /settings/active-page` has no `board_id`) → **#1247** (dashboard per-board active display)

#1185 ships the multi-board **foundation** (per-board settings + schema migration, per-board schedule storage/filtering by `board_id`, sidebar selector + CurrentBoardProvider, unified Schedule page). Full per-board driving is intentionally deferred to #1243/#1244/#1247. The single product decision for this release: ship the foundation now with driving to follow, or hold #1185 until those land.

Issue #1381, opened 2026-07-16
Migrate web UI to react-router 8 + vite 8


jeffredodd:
## Context
Dependabot's npm group PR #1366 originally bumped react-router 7.18→8.1, vite 7.3→8.1, and @vitejs/plugin-react 5→6. Those majors were backed out of that PR because they break CI; this issue tracks the actual migration.

## What breaks today
- **/schedule route reload-loops under RR8/vite8**: every schedule-related E2E and smoke test fails with `getByRole('heading', { name: 'Schedule' })` never appearing; the Playwright call logs show the page repeatedly re-navigating to `/schedule` (`commit → domcontentloaded → load` cycling). The signature matches React Router's "reload document on failed lazy route-chunk import" behavior — likely a vite 8 chunk-manifest/naming change or an RR8 route-module breaking change scoped to that route. All other routes passed. Evidence: E2E run https://github.com/Fiestaboard/FiestaBoard/actions/runs/28912070701/job/85771636940 and smoke run https://github.com/Fiestaboard/FiestaBoard/actions/runs/28912070698/job/85771367236 (Playwright traces in the `playwright-results` artifact).
- **Node engine floor**: react-router 8 requires Node ≥ 22.22. The Docker image is on Node 26, but several CI jobs (`.github/workflows/ci.yml`) still use `node-version: '20'` and emit `EBADENGINE` warnings; these should be bumped to 22+ as part of the migration.

## Suggested plan
1. Bump CI `node-version: '20'` → `'22'` (or 26 to match the Dockerfile).
2. Recreate the bump on a branch (`react-router@8`, `@react-router/*@8`, `vite@8`, `@vitejs/plugin-react@6`), rebuild the dev container, and reproduce the `/schedule` loop locally (devtools should show the failed chunk import or route-module error).
3. Follow the react-router v8 migration guide for the schedule route module; check vite 8 build-output changes.
4. Re-enable the majors for Dependabot afterwards.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1382, opened 2026-07-16
chore(deps): ignore react-router/vite major bumps until migration

jeffredodd:
## Summary
Adds Dependabot `ignore` rules for major updates of `react-router`, `@react-router/*`, `vite`, and `@vitejs/plugin-react`.

The v8 majors were backed out of #1366 because react-router 8 breaks the `/schedule` route with an endless reload loop and requires Node ≥ 22.22 (CI runs Node 20 in several jobs). Without these rules, next week's grouped npm PR would re-propose the same broken majors. The migration is tracked in #1381 — remove these entries when it lands.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,6 +19,17 @@ updates:
       npm-dependencies:
         patterns:
           - "*"
+    ignore:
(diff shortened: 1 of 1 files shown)


jeffredodd:
@claude Can you log an issue for this migration we need to do?

PR #1461, opened 2026-07-30
feat(docs-site): permanent redirects for retired doc URLs

jeffredodd:
## Why

The docs site shows a steady stream of "Page Not Found" hits in GA. Mining the full git history of `docs-site/` found the cause: it's not page renames (there have been literally zero file renames or slug changes since the site launched in Feb) — it's **retired URL prefixes**:

1. **Dropped version snapshots** — deploys build only the 12 most recent minors since the OOM fix (#: `DEPLOY_VERSION_CAP`), but all ~59 older snapshots (`/docs/2.5/…` through `/docs/8.0/…`) were live and search-indexed before the cap landed. Every one of those URLs now 404s.
2. **`/docs/next/*`** — the "Next" version was removed in #424 (Mar 2026); its URLs were indexed while live.
3. **`/docs/setup/split-flap-display-software`** — deleted in #419, content folded into intro/quick-start.

## What

GitHub Pages is static hosting, so a real HTTP 301 isn't possible. This adds `@docusaurus/plugin-client-redirects` (pinned to the Docusaurus version, 3.9.2), which emits a stub HTML page at each old URL with an instant `meta http-equiv="refresh"` + `rel="canonical"` to the new URL — the combination search engines treat as a permanent redirect. Visitors never see the 404 page.

The redirect map is one `createRedirects` rule: every current docs route gets from-paths for each retired prefix (`versions.slice(DEPLOY_VERSION_CAP)` + `"next"`). This is **self-maintaining** — when a future release pushes another snapshot off the cap, its redirects appear on the next deploy with no config change. The one deleted page gets an explicit entry.

## Verification

- Cross-checked all 59 dropped snapshots against the latest snapshot: **every page in every retired version has a same-path page in current docs** (zero orphans), so same-path redirecting is exact.
- Local `DOCS_PR_MODE=1` build: 3,722 stubs generated; spot-checked `/docs/2.5/intro`, `/docs/6.5/setup/quick-start`, `/docs/next/plugins/weather`, `/docs/7.3/features/schedule`, and the deleted setup page — all carry correct refresh target + canonical.
- No collisions possible with served versions: retired prefixes and built versions are disjoint by construction, and served-version routes (`/docs/8.11/…`) are excluded from redirect generation.
- `prettier --check` clean; the `tsc`/eslint complaints in `docs-site` (stats.tsx lucide cast, config import sort) are pre-existing and untouched.


jeffredodd:
Second commit extends this based on follow-up discussion:

- **`/docs/latest/<page>` evergreen alias** — always forwards to the current docs at `/docs/<page>` (bare `/docs/latest` → `/docs/intro`). `/docs/` itself remains the canonical home of current docs; the alias is for humans who expect a "latest" URL.
- **Version dropdown labels the newest version "8.12 (latest)"** so it's clear `/docs/` is always current.
- **All non-latest built versions get `noIndex`** — search engines consolidate on current docs, and noIndex pages drop out of the sitemap automatically (verified: full deploy-mode build sitemap contains 0 versioned URLs, 62 current-docs URLs).
- robots.txt deliberately stays allow-all: a `Disallow` on old versions would prevent crawlers from ever seeing the redirect stubs and noIndex tags they need to clean up the index.

Verified with a full 12-version deploy-mode build (8GB heap, same as docs.yml) *and* a `DOCS_PR_MODE=1` build: noIndex meta present on 8.11/8.1 pages and absent on latest, all alias stubs point to correct targets, dropped-version stubs unchanged.

jeffredodd:
Third commit closes the gaps found in an adversarial self-review of the redirect map:

- **Bare `/docs` 404ed** (no doc has slug `/`) — now redirects to `/docs/intro`, same as docusaurus.io does for its own bare `/docs`. Ditto the bare form of every redirected prefix (`/docs/latest`, `/docs/next`, `/docs/<retired-version>`).
- **Bare roots of served older versions** (`/docs/8.11`) forward to that version's home page. These entries are derived from `builtVersions`, so they only exist in deploy builds where their target routes exist — PR-mode builds stay valid.
- **Week-one category URLs**: the site launched (2026-02-15) with an autogenerated sidebar, which generated `/docs/category/{setup,plugins,features,development,deployment,reference}` index pages; #295 (2026-02-22) switched to explicit sidebars and killed those routes. Each now forwards to its section's entry page.

Verification: full 12-version deploy build — 16/16 redirect target checks pass and every target resolves to a real page (not another stub); PR-mode build green; prettier clean.

Review context: the underlying mechanism was cross-checked against Google's redirect docs (instant meta refresh = treated as a permanent redirect) and against how docusaurus.io, react-native-website, and Consensys-style doc sites handle versioned-docs SEO. Two known non-issues left out deliberately: the pre-existing trailing-slash canonical mismatch on GH Pages (fixing it changes every canonical/sitemap URL — separate decision), and on-site local search still surfacing old versions (noIndex doesn't affect the local search index; there are ignore options if wanted later).

PR #1504, opened 2026-08-06
docs-site: render the site's own React on FiestaUI primitives + add the drift guard

jeffredodd:
**Stack 3 of 3.** Based on #1503, which is based on #1502. Review those first — this diff is only the primitives layer on top.

## The problem

`docs-site/src` was **213 raw HTML elements** deep — 109 `div`, 25 `p`, 20 `span`, 14 `button`, 8 `h2`, 3 `ul` — with not one design-system primitive in the site's own React, and `docs-site/eslint.config.mjs` had no restricted-element rule at all. So even after Tiers 1 and 2 fixed what was on screen, nothing stopped it drifting straight back.

## What changed

`src/components` and `src/pages` now render `Box` / `Flex` / `Text` / `Heading` / `List` / `Button`, and the `react/forbid-elements` rule from `web/eslint.config.mjs` is wired up to hold the line. Where a primitive takes over layout the CSS module was doing, the duplicated declarations are deleted rather than left to fight the utility classes — `.heroButtons`, `.currentCard`, `.chipGrid`.

The migration surfaced four more components the site had hand-rolled rather than imported:

| Hand-rolled | Now |
|---|---|
| `<div class="emptyState"><p>No plugins match…` | `EmptyState` (plugin directory + "plugin not found") |
| `<input type="search">` | `Input` |
| `<div class="skeleton"><div/><div/><div/>` ×3 | `Skeleton` |
| `<span class="categoryBadge">`, `<div class="spotlightBadge">` | `Badge` |

## Two deliberate exemptions

Both documented at the point they apply, so the next person doesn't have to guess:

- **`src/theme/**`** — swizzled Docusaurus components have to mirror upstream's markup and Infima class names (`.menu__list-item`, `.pagination-nav__link`, `.clean-btn`) to keep their behaviour. Forcing primitives there would break the theme rather than align it. Those files already use DS components wherever the markup is genuinely ours (see #1502, #1503).
- **The react-markdown renderer map in `plugins/detail.tsx`** — those are markdown *output* nodes, so the prose ones stay raw for exactly the reason the docs' own MDX does in #1502: `Text`/`Heading`/`List` are sized for the app's dense chrome, and README bodies need the long-form scale `detail.module.css` defines. Its **table** nodes are the exception and now map onto the DS `Table` set, matching `src/theme/MDXComponents.ts`.

`h1` is also absent from the forbid list. The web app sends it to `PageHeader`, but that's app chrome, and `Heading` covers h2–h4 only — a docs page's single h1 has no primitive to move to. Worth fixing in FiestaUI rather than papering over here.

## Verification

- `npm run typecheck`, `npm run lint` (0 violations, guard active), `npm run format:check` clean
- `DOCS_PR_MODE=1 npm run build` succeeds
- Rendered and inspected every migrated surface in the browser: homepage hero + feature/highlight/showcase/plugin sections, plugin directory (search, filters, board toggle, cards, and the empty state with a no-match query), stats page (stat strip, bar chart, spotlight, recently-added/updated lists), plugin detail (header, board toggle, README with a DS-rendered config table), and the screenshot toggle + lightbox in the docs — including the lightbox's dark-overlay footer, where the DS `Button` had to keep the contextual white-tinted styling

## Follow-up worth doing separately

`BoardScreenshot` and `ThemedScreenshot` each hand-roll a lightbox: a `div role="dialog" aria-modal="true"` with an Escape listener and a manual `body.overflow` lock, but no focus trap and no focus restore. FiestaUI's `Dialog` gives all three. I left it out to keep this PR to the migration, but it's a genuine a11y gap in two components used across many docs pages.

### DIFF
--- a/docs-site/eslint.config.mjs
+++ b/docs-site/eslint.config.mjs
@@ -60,5 +60,58 @@ export default [
       "react-hooks/exhaustive-deps": "warn",
     },
   },
+  // Design-system enforcement, mirroring web/eslint.config.mjs: the site's own
+  // React renders @fiestaboard/ui components, not raw HTML. Allowlisted leaves
+  // (svg, canvas, iframe, img, br, em, small, kbd, pre, figure, figcaption, dl,
+  // dt, dd, input, hr, blockquote) are simply not listed here.
+  //
+  // Scoped to src/components and src/pages on purpose. src/theme holds swizzled
+  // Docusaurus components, which must mirror upstream's markup and Infima class
+  // names to keep their behaviour (`.menu__list-item`, `.pagination-nav__link`,
+  // `.clean-btn`); forcing primitives there would break the theme rather than
+  // align it. Those files use DS components where the markup is genuinely ours.
+  {
+    files: ["src/components/**/*.tsx", "src/pages/**/*.tsx"],
+    rules: {
+      "react/forbid-elements": [
+        "error",
+        {
+          forbid: [
+            { element: "div", message: "Use Flex/Stack/Grid for layout, or Box (@fiestaboard/ui)" },
+            { element: "span", message: 'Use Text as="span" (@fiestaboard/ui)' },
+            { element: "p", message: "Use Text (@fiestaboard/ui)" },
+            // `h1` is intentionally absent. The web app forbids it in favour of
+            // `PageHeader`, but that is app chrome; FiestaUI's `Heading` covers
+            // h2-h4 only, so a docs/marketing page's single h1 has no primitive
+            // to move to and keeps its own element.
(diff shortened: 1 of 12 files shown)


jeffredodd:
Rebased onto #1505.

The original base (`feat-docs-fiestaui-chrome`) became unreachable from `main`: #1503 merged into the Tier 1 branch rather than `main`, and #1502 squash-merged Tier 1 separately, so the Tier 2 commits never landed. #1505 re-opens that work against current `main` and this PR now stacks on it. Diff is unchanged — still just the Tier 3 primitives migration + lint guard.

Issue #1436, opened 2026-07-25
[Docs] add Note previews


roblesi:
All the screenshots are flagship, we should have both in the docs.

PR #1518, opened 2026-08-08
feat(plugins): plugin-previews.json seed for all 51 registry plugins + daily sync

jeffredodd:
**PR 3 of 4** of the board-previews project (refs #1436). Stacked on #1509 — when that merges this retargets to `main` automatically.

## What

- **`plugin-previews.json`** — the central seed: `teaser` + flagship **and** note literal previews for all 51 registry plugins, in the exact contract #1509 defines. The docs site renders this when a plugin's own manifest doesn't declare the fields yet.
- **`scripts/sync_plugin_previews.py`** — the merge that makes "seed now, backfill later" safe: manifest `teaser`/`previews` win, seed entries are the fallback, and entries are removed only when a plugin leaves the registry. An invalid or half-adopted manifest declaration warns and keeps the existing entry — adopting can only upgrade, never regress.
- **`scheduled-maintenance.yml`** — the sync runs in the existing daily plugin-stats job and commits alongside `plugin-stats.json`.
- **`tests/test_sync_plugin_previews.py`** — 21 tests: seed integrity (every registry plugin present, teaser valid, previews valid, flagship+note coverage — the #1436 requirement, now guarded in CI) and the resolution-order rules.

## How the seed was generated

- Flagship rows port the hand-authored `PLUGIN_DISPLAYS` entries from `web/tests/generate-screenshots.spec.ts` where they exist (25 plugins; `baywheels` → `lyft_bike_share` id mapped).
- Everything else — including **every note board** — was rendered through the real `TemplateEngine` (`render_lines` with each demo's `line_metadata`) using the plugin's own `demo` templates and declared variable `example` values, then hand-polished where examples were placeholders (`hacker_news` `???`s, `sun_art`/`visual_clock`/`white_noise`/`generative_ai_art` note boards, `disney_parks_times` note) or truncated mid-word (`volcano`, `word_of_day`, `spacecraft_launches`, `health`, `last_fm`, `stocks` note).
- All 51 teasers are hand-authored, ≤15 tiles.
- The whole file validates against `src/plugins/previews.py` — zero errors, and the integrity tests keep it that way.

The sync script imports only stdlib + `src.plugins.previews` (itself stdlib-only), so it runs on a bare runner with no pip install — same pattern as `fetch_plugin_stats.py`.

## Verified

- `pytest tests/test_sync_plugin_previews.py tests/test_plugin_previews.py tests/test_manifest_previews.py tests/test_plugin_validation.py` — 168 passed (in the dev container)
- `ruff check` + `ruff format --check` clean at the pinned 0.16.1

Per the design doc, `PLUGIN_DISPLAYS` deletion waits until PR 4 lands and the screenshot pass becomes dead code.

### DIFF
--- a/.github/workflows/scheduled-maintenance.yml
+++ b/.github/workflows/scheduled-maintenance.yml
@@ -7,7 +7,9 @@ name: 'Scheduled: Maintenance'
 #
 # Jobs:
 #   - plugin-stats   : daily refresh of docs-site/static/plugin-stats.json
-#                      (was .github/workflows/plugin-stats.yml)
+#                      (was .github/workflows/plugin-stats.yml) and of
+#                      plugin-previews.json (manifest teaser/previews win,
+#                      seed entries are the fallback)
 #   - security-scan  : weekly bandit + pip-audit sweep of plugin-registry
 #                      repos, opens/updates an issue on findings
 #                      (was .github/workflows/registry-scan.yml)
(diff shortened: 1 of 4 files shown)


jeffredodd:
Per maintainer feedback the note previews were too plain next to the flagship boards — all 47 text-only note boards in the seed were redesigned with color: full banners (weather, air quality, tornado warning), plugin-hued title accent chips, data-colored values (on-time green / delay red / UV orange), and the existing art gradients, with deliberate variety so the note column doesn't read as one template. The 50 plugin-repo backfill PR branches were refreshed to match (46 updated, 4 already colorful), and the bundled plugins on #1509 got the same treatment.

PR #1520, opened 2026-08-08
fix(ci): stop integration tests from inflating plugin clone stats

jeffredodd:
## Summary
- The docs stats page (fiestaboard.app/stats) showed Dad Jokes at 678 unique cloners / 2,302 clones in 14 days and crowned it "Most popular" — but the traffic was our own CI. `tests/test_plugin_install_integration.py` really cloned `fiestaboard-plugin--dad-jokes` (~8×/run) and `--star-trek-quotes` (~3×/run) on every `python-tests` job, and each Actions runner IP counts as a new "unique cloner" in the GitHub Traffic API. The daily traffic breakdown is bursty and weekday-shaped (456 clones on Aug 2, 408 on Jul 25), unlike the steady ~300-unique fleet baseline of genuinely popular plugins like weather/stocks.
- Rewrite the integration tests to run the same real-git install/update/uninstall flows against local `file://` fixture repos — hermetic, no network flake, and zero pollution of the plugin repos' traffic stats.

## Changes
- `tests/test_plugin_install_integration.py`: local fixture repo (`fiestaboard-plugin--ci-fixture`) built per-test with git init/commit; all 13 tests exercise the real `git init`+`fetch`+`reset` install path, update path, and uninstall against it. The https-only URL validation is bypassed via monkeypatch (it stays unit-tested in `test_plugin_sources.py`). The update test now asserts a new upstream commit actually arrives instead of just comparing history length.
- `.github/workflows/ci.yml`: updated the step comment to record why the tests are hermetic.
- `docs-site/src/pages/stats.tsx`: honest labels — "unique cloners" instead of "unique installs", and the footer notes counts include automated traffic (fleet auto-updates, weekly security scan).

Note: GitHub's traffic window is a rolling 14 days, so the inflated Dad Jokes / Star Trek Quotes numbers will decay over the two weeks after this merges — they can't be corrected retroactively.

## Test plan
- [x] `pytest -m integration tests/test_plugin_install_integration.py -v` in a container from the project image — 13 passed in 0.20s, fully offline
- [x] `ruff check` + `ruff format --check` on the test file — clean
- [x] `prettier --check src/pages/stats.tsx` (prettier 3.4.2, docs-site config) — clean

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -203,8 +203,10 @@ jobs:
 
     - name: Run plugin install integration tests
       run: |
-        # Real git clone/pull/uninstall against live Fiestaboard plugin repos on GitHub.
-        # Requires network access and git binary (both available on ubuntu-latest).
+        # Real git install/update/uninstall flows against local file:// fixture
+        # repos — hermetic on purpose. Cloning the live plugin repos from here
+        # made every CI run count as a "unique cloner" in their GitHub traffic
+        # stats, inflating fiestaboard.app/stats (dad-jokes, star-trek-quotes).
         pytest -m integration tests/test_plugin_install_integration.py -v
       env:
         BOARD_READ_WRITE_KEY: test_key
--- a/docs-site/src/pages/stats.tsx
+++ b/docs-site/src/pages/stats.tsx
@@ -16,8 +16,10 @@ interface PluginStat {
   version: string | null;
   created_at: string | null;
   updated_at: string | null;
-  clones_14d_count: number;
-  clones_14d_uniques: number;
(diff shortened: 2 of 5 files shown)


jeffredodd:
Added a second commit: `fetch_plugin_stats.py` hardcoded the `Fiestaboard/` org, so `airport_departures` (community-hosted at `breedloj/…`) 404'd on every API call and showed `version: null` / 0 clones. Now the owner is derived from the registry URL — version and dates populate. Clone traffic for community repos is unobtainable (GitHub's traffic API requires push access), so it's reported as `null` and the stats page excludes those plugins from the ranking/totals with an explanatory note instead of showing a fake 0.

PR #1526, opened 2026-08-08
feat(schedule): Order schedule entries by next occurrence

agrahamg:
## Description
The Schedule Entries list rendered in creation order, which made it hard to read off when the board actually changes. 

This PR changes to sort chronologically instead: weekly entries first (the everyday baseline, ordered by start time), then annual and one-off entries merged and ordered by their next occurrence, so that section reads as a queue of what is coming up. One-offs whose window has already passed sink to the bottom, most recent first.

Sun-based entries sort on their server-resolved time, so a sunrise entry lands where it actually fires rather than at its stored fallback. The reference date is injectable so the ordering stays testable as days pass.

### Before
Screenshot 2026-08-08 at 12 40 42 PM
### After
Screenshot 2026-08-08 at 12 41 03 PM

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] New plugin
- [ ] Documentation update
- [ ] Refactor / maintenance

## Checklist

- [X] I have tested my changes locally
- [X] I have added/updated tests where applicable
- [X] I have updated documentation where applicable
- [X] My code follows the project's coding standards

### DIFF
--- a/web/src/__tests__/schedule-sort.test.ts
+++ b/web/src/__tests__/schedule-sort.test.ts
@@ -1,0 +1,148 @@
+import { describe, expect, it } from "vitest";
+
+import type { ScheduleEntry } from "@/lib/api";
+import { sortSchedulesByStart } from "@/lib/schedule-sort";
+
+// Pinned reference point so "next occurrence" ordering is deterministic.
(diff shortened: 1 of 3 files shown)


roblesi:
What a great feature, thanks for the PR!

jeffredodd (approved):

Issue #856, opened 2026-06-02
Calendar Plugin Page can not be Manually changed


cordell25:
**Describe the bug**
When the Calendar plugin triggers its default page to be displayed for an upcoming event, this page is unable to be overwritten.

**To Reproduce**
1. Setup the Calendar plugin
2. Wait for the next event to trigger a display on the Vestaboard
3. Attempt to send a Page to the Vestaboard manually from the Home Screen using "Change Page"
4. The calendar display is Not replaced, the "Change Page" request appears to be ignored

**Expected behavior**
Current display is replaced by the Page selected in step 3

**Environment**
- FiestaBoard version: [6.10.0]
- Board connection mode: [local]
- Hardware: [Raspberry Pi 3]

**Additional context**
I believe there is some sort of priority issue happening, the Manual Change Page needs to have the highest priority so that the user can always override whatever is displayed on their board, regardless of how it got there.

jeffredodd:
PR #855 includes a fix for this. The root cause was per-loop-tick re-emission of the same `trigger_id` from `calendar_sub` (and any other trigger plugin that re-fires while a condition holds): the user's "Change Page" hit `PUT /settings/active-page` and went through, but the next tick `activate_trigger` happily replaced the user's content with the same trigger again.

The fix adds per-`trigger_id` suppression to `TriggerService`. When `PUT /settings/active-page` runs, it now calls `dismiss_active_for_user_override()` which dismisses *and* blacklists every active trigger's id until its natural duration would have ended. New triggers from different events (different ids) still fire normally — only the dismissed ones stay quiet.

Verified locally against your scenario: with a calendar countdown trigger active, manually setting an Active Page now keeps that page on the board, and re-runs of `POST /triggers/check` no longer flip the trigger back on. 43/43 trigger unit tests pass.

jeffredodd:
@cordell25 let me know if this fixes the issue for you. I think this really comes down to the fact we've not tested triggers well enough so I'm not expecting this to be the only fixed need. Let me know and thanks again!

PR #1561, opened 2026-08-11
fix(mcp): route set_active_page and set_schedule_mode at services that exist

### LINKED ISSUE #1559: MCP `set_active_page` always fails: calls non-existent `ConfigManager.set_active_page()`
**Describe the bug**
The MCP tool `set_active_page` can never succeed. It calls `cm.set_active_page(page_id)` on `ConfigManager`, which has no such method. Every call raises `AttributeError`, is caught by the tool's own `except`, and is returned to the client as an error string. The call is a complete no-op... nothing is written and the board does not change.

**To Reproduce**
1. Connect any MCP client to `/api/mcp/`.
2. Obtain a valid page ID from `create_page(...)` or `list_pages()`.
3. Call `set_active_page(page_id)`.

**Expected behavior**
The page becomes active and is pushed to the board, as `POST /api/active-page` does.

## Actual

```json
{"status":"error",
 "error":"Error setting active page: 'ConfigManager' object has no attribute 'set_active_page'"}
```

Reproduces with a valid, existing page IDa including one that is *already* the active page. Not an input-validation problem.

## Root cause

`src/mcp_server.py:958-975`:

```python
@mcp.tool()
def set_active_page(page_id: str) -> dict[str, Any]:
    try:
        from .config_manager import get_config_manager

        cm = get_config_manager()
        cm.set_active_page(page_id)          # <-- ConfigManager has no such method
        return _ok(f"Active page set to '{page_id}'.", page_id=page_id)
    except Exception as exc:
        return _err(f"Error setting active page: {exc}")
```

The real API is `SettingsService.set_active_page_id(page_id, board_id=None)` at `src/settings/service.py:1009`.

`mcp_server.py:972` appears to be the only call site using the `ConfigManager` form. All others go through the settings service:

- `src/api_server.py:6001,6003,6178`
- `src/mqtt/commands.py:195,198,368,384`
- `src/main.py:676,704`
- `src/pages/service.py:244,267`

This looks like a rename of the settings API that missed one call site.

## Evidence that nothing is written

`data/settings.json` mtime before and after a `set_active_page` call:

```
before: 2026-08-10 23:13:06.023494036 +0000
after:  2026-08-10 23:13:06.023494036 +0000
```

Identical. The failure precedes any persistence.

## Suggested fix

Route the MCP tool through the same path as the REST endpoint (`api_server.py:5938`) rather than calling the settings service directly, so it also picks up `dismiss_active_for_user_override()`, board resolution, and the immediate send. Calling `set_active_page_id()` alone would persist the change but may not push it to the display.

Minimal correction to the failing line:

```python
from .settings.service import get_settings_service

get_settings_service().set_active_page_id(page_id)
```

## Note: `create_page` is unaffected

`create_page()` does correctly activate and display a newly created page, so creating pages via MCP works end to end. Only switching to an existing page is broken. (Not determined: whether `create_page` auto-activates unconditionally or only when no other page exists.)

**Environment**
- OS: Ubuntu 22.04.2, Docker 24.0.5
- Deployment: Docker Compose via `docker-compose.hub.yml`, x86_64
- FiestaBoard **8.25.5** (`fiestaboard/fiestaboard:latest`, pulled 2026-08-10)
- MCP client: Claude Desktop via `mcp-remote`
- Board connection mode: Local
- Board: Vestaboard Flagship, firmware 4.5.4, Local API
- Hardware: PC - Ryzen 7 mini PC from Beelink

**Additional context**

jeffredodd:
Fixes #1559.

## The bug

`set_active_page` called `ConfigManager.set_active_page()` — a method that has never existed. The tool caught its own `AttributeError` and returned it to the client as an error string, so it was a complete no-op: nothing written, board unchanged, in every release.

Hardening the tests to catch it found a **second, unreported instance of the same defect**:

| Tool | Called | Actually lives on |
|---|---|---|
| `set_active_page` | `ConfigManager.set_active_page()` | `SettingsService.set_active_page_id()` |
| `set_schedule_mode` | `ScheduleService.set_schedule_enabled()` | `SettingsService.set_schedule_enabled()` |

This is one defect, not two: the settings API moved to `SettingsService` and these two call sites were left behind. Every other caller — REST (`api_server.py`), MQTT (`mqtt/commands.py`), the main loop, the page service — had been updated. MCP schedule mode has never worked and nobody had reported it.

## The fix

**`set_active_page`** now delegates to the `PUT /settings/active-page` handler instead of reimplementing it. Selecting a page is more than a settings write — it validates the ref, enforces page↔board size compatibility, dismisses active plugin triggers so the choice sticks (#856), and renders to the board. Delegating removes the drift that caused this in the first place. The tool became `async` to await the handler, and now returns `sent_to_board` / `paused` so a client can tell the user *why* the board hasn't changed yet (board paused, or waiting for the next display refresh).

**`set_schedule_mode`** routes through `SettingsService`, matching the REST endpoint exactly.

## Why no test caught it

There *was* a test for `set_active_page`, and it passed:

```python
mock_cm = MagicMock()
...
mock_cm.set_active_page.assert_called_once_with("page-001")   # phantom method
```

`MagicMock()` conjures any attribute on access. The test asserted that production code called a method that doesn't exist — and agreed with it. Every service mock in the file had this problem, so none of the ~60 tool tests could catch a miswired call.

## Testing

**1. Autospec fixtures.** All six service fixtures in `test_mcp_server.py` now use `create_autospec(RealClass, instance=True)`, so a phantom method raises instead of passing. All 93 existing tests still pass — nothing else was hiding in the paths tests already execute.

**2. Behavioral tests** for both tools, asserting real outcomes rather than mock calls: the selection persists, the page is pushed to the board, plugin triggers are dismissed, and an unknown page id errors *without* persisting.

**3. `tests/test_service_wiring.py` (new).** A contract test that parses `src/mcp_server.py` and `src/mqtt/commands.py`, resolves every `svc = get_x_service(); svc.method()` call against the real class behind the getter, and fails if the method doesn't exist. This is what found the `set_schedule_mode` bug — it covers all 39 MCP tools whether or not any test calls them. MQTT is included because it's the other surface with the same lazy-import-and-swallow-exceptions shape; it scans clean.

The analyzer handles the three traps that would make this check quietly useless:
- **nested scoping** — all 39 tools live inside `_build_mcp_server`, so a naive `ast.walk` pools their locals into one namespace where `svc` means whatever the last tool assigned
- **chained calls** — `get_settings_service().method()`, not just local bindings
- **`Optional` returns** — `get_service() -> DisplayService | None` must resolve to `DisplayService`, not `Union`

## Evidence the tests aren't vacuous

The wiring test carries meta-tests asserting the analyzer flags a known-bad call, accepts a known-good one, and doesn't leak locals between sibling functions — plus a floor on resolved-call count (43 and 21 actual, 35/15 floors) so it fails loudly if it ever goes blind instead of passing on zero checks.

I reintroduced both bugs in `mcp_server.py` and re-ran the suite:

```
FAILED tests/test_service_wiring.py::test_service_calls_resolve_to_real_methods[src/mcp_server.py]
E   AssertionError: Calls to methods that do not exist:
E     src/mcp_server.py:978 in set_active_page(): get_config_manager().set_active_page
E         — ConfigManager has no attribute 'set_active_page'
E     src/mcp_server.py:1014 in set_schedule_mode(): svc.set_schedule_enabled
E         — ScheduleService has no attribute 'set_schedule_enabled'

FAILED tests/test_mcp_server.py::TestSetActivePage::test_persists_selection_via_settings_service
FAILED tests/test_mcp_server.py::TestSetActivePage::test_pushes_the_page_to_the_board
FAILED tests/test_mcp_server.py::TestSetActivePage::test_dismisses_plugin_triggers_so_the_choice_sticks
FAILED tests/test_mcp_server.py::TestSetActivePage::test_unknown_page_reports_error_and_persists_nothing
FAILED tests/test_mcp_server.py::TestSetScheduleMode::test_enable_persists_via_settings_service
FAILED tests/test_mcp_server.py::TestSetScheduleMode::test_disable_persists_via_settings_service

7 failed, 93 passed
```

Then restored and confirmed green: 296 passed across `test_mcp_server`, `test_service_wiring`, the MCP auth suites, `test_api_coverage`, and `tests/contract`. `ruff check` clean over `src/` and the changed tests; `ruff format --check` clean.

## End-to-end verification

Ran against real services with no mocks (containerized, isolated data dir):

```
before         : b7d08ab8-…
set_active_page: {'status': 'success', 'message': "Active page set to '94ccceef-…'.
                  It will appear on the board on the next display refresh.",
                  'sent_to_board': False, 'paused': False, 'warnings': []}
after          : 94ccceef-…
bad page       : {'status': 'error', 'error': 'Error setting active page: Page not found: does-not-exist'}
set_schedule_mode(True)  -> persisted=True
set_schedule_mode(False) -> persisted=False
```

The active page persists to `settings.json`, a bad id returns a clean error without clobbering the current selection, and schedule mode persists both directions. (`sent_to_board: False` because no board is configured in that bare environment — the message explains it.)

### DIFF
--- a/src/mcp_server.py
+++ b/src/mcp_server.py
@@ -957,7 +957,7 @@ def _build_mcp_server() -> Any:
             return _err(str(exc))
 
     @mcp.tool()
-    def set_active_page(page_id: str) -> dict[str, Any]:
+    async def set_active_page(page_id: str) -> dict[str, Any]:
         """Set which page is currently shown on the FiestaBoard display.
 
         This immediately changes what's visible on the board.
@@ -965,14 +965,34 @@ def _build_mcp_server() -> Any:
         Args:
             page_id: The page or collection ID to display (from list_pages() or list_collections()).
         """
-        try:
-            from .config_manager import get_config_manager
+        # Delegate to the REST handler rather than reimplementing it. Selecting
+        # a page is more than a settings write: it validates the ref, enforces
+        # page<->board size compatibility, dismisses active plugin triggers so
+        # the choice sticks (#856), and renders to the board. Issue #1559 was
+        # this tool going its own way and calling a method that never existed.
+        from fastapi import HTTPException
 
-            cm = get_config_manager()
-            cm.set_active_page(page_id)
-            return _ok(f"Active page set to '{page_id}'.", page_id=page_id)
+        from .api_server import set_active_page as _rest_set_active_page
+
+        try:
+            response = await _rest_set_active_page({"page_id": page_id})
+        except HTTPException as exc:
+            return _err(f"Error setting active page: {exc.detail}")
         except Exception as exc:
             return _err(f"Error setting active page: {exc}")
+
+        message = f"Active page set to '{page_id}'."
+        if response.get("paused"):
+            message += " The board is paused, so it will appear when you resume it."
+        elif not response.get("sent_to_board"):
+            message += " It will appear on the board on the next display refresh."
+        return _ok(
+            message,
+            page_id=page_id,
+            sent_to_board=bool(response.get("sent_to_board")),
+            paused=bool(response.get("paused")),
+            warnings=response.get("warnings", []),
+        )
 
     @mcp.tool()
     def set_schedule_mode(enabled: bool) -> dict[str, Any]:
@@ -985,9 +1005,11 @@ def _build_mcp_server() -> Any:
             enabled: True to enable schedule-based display, False to disable.
         """
         try:
-            from .schedules.service import get_schedule_service
+            # Schedule mode is a settings flag (per-board since #1244), not
+            # something ScheduleService owns — same mixup as issue #1559.
+            from .settings.service import get_settings_service
 
-            svc = get_schedule_service()
+            svc = get_settings_service()
             svc.set_schedule_enabled(enabled)
             state = "enabled" if enabled else "disabled"
             return _ok(f"Schedule mode {state}.", enabled=enabled)
--- a/tests/test_mcp_server.py
+++ b/tests/test_mcp_server.py
@@ -17,8 +17,9 @@ Strategy
 
 from __future__ import annotations
 
+from types import SimpleNamespace
 from typing import Any
-from unittest.mock import MagicMock, patch
+from unittest.mock import MagicMock, create_autospec, patch
 
 import pytest
 
@@ -49,7 +50,9 @@ def _call_tool(mcp_instance: Any, tool_name: str, **kwargs: Any) -> Any:
         raise KeyError(f"Tool '{tool_name}' not registered. Available: {list(mgr._tools)}")
     result = tool.fn(**kwargs)
     if asyncio.iscoroutine(result):
-        result = asyncio.get_event_loop().run_until_complete(result)
+        # asyncio.run(), not get_event_loop(): 3.14 raises RuntimeError when
+        # there is no current loop rather than creating one implicitly.
+        result = asyncio.run(result)
     return result
 
 
@@ -68,8 +71,10 @@ def mcp():
 
 @pytest.fixture
 def mock_registry():
-    """Minimal mock plugin registry."""
-    registry = MagicMock()
+    """Minimal mock plugin registry, specced against the real class."""
+    from src.plugins.registry import PluginRegistry
+
+    registry = create_autospec(PluginRegistry, instance=True)
     registry.list_plugins.return_value = [
         {
             "id": "openweather",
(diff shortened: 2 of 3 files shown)


jeffredodd:
Note for reviewers: the first CI run on this branch showed `Platform Tests` failing on `tests/test_api_coverage.py::TestSendWelcomeMessage::test_welcome_success` (`assert 500 == 200`). That is not related to this change — a re-run went green with no code changes.

I chased it down rather than just re-running: the Python suite shares one `data/config.json` across all pytest-xdist workers, and `ConfigManager._save_internal()` writes through a **fixed** `config.json.tmp` before renaming. Concurrent writers collide, giving either a lost write or `FileNotFoundError: 'data/config.json.tmp' -> 'data/config.json'`. `main` reproduces the same failure family independently of this branch at `-n 16`, so it is pre-existing.

Filed as #1564 with a repro and a suggested fix. Deliberately not fixing it here — it is unrelated to the MCP wiring and deserves its own review.

Verification for this branch specifically: full suite at CI's worker count (`-n 4`), five consecutive runs, `4745 passed` every time.

jeffredodd:
Correction to my note above: the fix for that flake is already in flight as **#1563** (`fix/config-save-tmp-race`), which landed while I was writing this up. It goes further than what I described — a shared `src/atomic_io.staging_path()` helper giving every JSON store under `data/` a per-process staging name, not just `config_manager`. I've closed my #1564 as a duplicate of it.

Nothing changes for this PR: the flake is pre-existing and unrelated, CI here is green.

PR #1589, opened 2026-08-12
feat(transitions): add global and per-page transition plugin pickers

jeffredodd:
## Why

From Discord: a community member enabled the Transitions beta, followed [`plugins/typewriter/docs/SETUP.md`](https://github.com/Fiestaboard/FiestaBoard/blob/main/plugins/typewriter/docs/SETUP.md), and could not find the UI it described.

They were right, and the docs were less wrong than they looked. Transition plugins were implemented end to end but **unreachable from any UI** — the only way to apply one was `PUT /settings/transitions` or `PUT /pages/{id}` by hand.

Per-page overrides were the more surprising gap. `Page.transition_strategy` is honored at send time (`src/main.py:774`, `:977`; `src/api_server.py:6025`, `:7707`) with fallback to the global default, is persisted, shareable, and clearable — it simply never had a control in the page editor.

## What changed

**Global picker** — `Settings → Behavior → Board Transitions` now lists installed transition plugins in a "Plugins (beta)" group beside the built-in strategies, saving `strategy` as `plugin:<id>`. Step Interval / Step Size are hidden for plugin strategies, since those are forwarded to the Local API and ignored by plugins.

**Per-page picker** — the page editor header gains a `Transition` dropdown: "Use global default", the built-in strategies, and (beta on) the installed plugins, plus a link to the Transition Lab. Clearing sends explicit `null` so the backend's nullable allowset actually drops the override rather than leaving it sticky.

Neither picker discards a setting it can't resolve: a `plugin:<id>` whose plugin is uninstalled, or whose beta flag is off, still renders under its raw id instead of silently reading as "Use global default".

**Integrations** — transition plugins no longer show an enable/disable toggle. It was a no-op: `get_transition_plugin()` (`src/plugins/registry.py:618`) deliberately ignores the enabled flag, so disabling Typewriter still ran Typewriter. They're badged "Transition" instead. `plugin_type` is now returned by `list_plugins()` and `GET /plugins/{plugin_id}`.

**Docs** — built-in strategies are Vestaboard **Local API** features; transition plugins are host-driven full-grid sends that work on **any** connection. The docs conflated the two, and `transitionLab.noPlugins` told users to "enable one on the Integrations page" — the exact step that doesn't exist. All four plugin READMEs + SETUP guides corrected, `TRANSITION_PLUGIN_DEVELOPMENT.md` updated, and a new `docs-site/docs/features/transitions.md`, since the site documented the feature nowhere at all.

## Test evidence

Every test was written before its implementation and proven non-vacuous by breaking the specific production behavior and recording the failure. Representative:

```
# omit the key when null, rather than sending explicit null
× sends an explicit null when clearing an existing override back to the global default
  AssertionError: expected { name: 'Weather Page', …(3) } to have property "transition_strategy"

# drop the plugin: prefix on save
  AssertionError: expected 'slot_machine' to be 'plugin:slot_machine'

# ungate the toggle so transitions render it again
× renders no enable toggle for a transition plugin
  expected document not to contain element, found <button ...

# remove plugin_type from the registry payload
E   KeyError: 'plugin_type'
```

- Python: **4679 passed**. Two files (`test_plugin_sources.py`, `test_plugin_install_integration.py`) fail only in the local test image, which has no `git` binary — unrelated and green in CI.
- Web: **1497 passed / 119 files**, including the zero-key-drift i18n guard after all 15 new keys were translated into the 13 non-English locales.
- Ruff, Prettier, ESLint, and `validate_plugins.py` (7/7) clean.
- `docs-site` verified with a real Docusaurus build; `onBrokenLinks: "throw"` passes.

## Follow-ups not in this PR

- The **Marketplace** tab still can't distinguish transition plugins — `plugin-registry.json` carries no `plugin_type` on any of its 51 entries, so it needs a `RegistryEntry` field plus a seed backfill.
- **Per-collection / per-schedule transitions**, the original feature request. Page-level is now the first rung of the hierarchy (page → global).

### DIFF
--- a/docs-site/docs/features/page-editor.md
+++ b/docs-site/docs/features/page-editor.md
@@ -158,6 +158,18 @@ The split-flap display supports colored tiles using special character codes. You
 
 See the [Color Guide](/docs/reference/color-guide) for detailed usage examples.
 
+## Page Transitions
+
+The editor's header toolbar has a **Transition** dropdown that controls how the board animates when it changes to this page.
+
+- **Use global default** - inherit the transition set in **Settings → Behavior → Board Transitions**
+- A **built-in strategy** (Wave, Drift, Curtain, Row, Diagonal, Random) - performed by the board itself, Local API only
+- A **transition plugin** (Typewriter, Simple Dissolve, and others) - animated by FiestaBoard, works on any connection, shown only when the Transition Plugins beta is enabled
+
+Anything other than **Use global default** becomes a page-level override that applies whenever this page is sent. Switching back to **Use global default** clears it.
+
+See [Transitions](/docs/features/transitions) for how the two kinds differ, what they require, and how to preview them in the Transition Lab.
+
 ## Page Layout
 
 The board display dimensions depend on the target device:
@@ -198,6 +210,7 @@ Pages that are referenced in schedules should be updated or removed from the sch
 
 ## Next Steps
 
+- [Transitions](/docs/features/transitions) - Choose how the board animates into this page
 - [Schedule Mode](/docs/features/schedule) - Automate when pages are displayed
 - [Plugins Overview](/docs/plugins/overview) - See available data sources for your pages
 - [Color Guide](/docs/reference/color-guide) - Learn about color formatting options
--- a/docs-site/docs/features/transitions.md
+++ b/docs-site/docs/features/transitions.md
@@ -1,0 +1,146 @@
+---
+sidebar_position: 6
+description: "Animate how your FiestaBoard changes between messages — built-in Vestaboard flip strategies and beta transition plugins like typewriter, dissolve, and slot machine."
+keywords: [FiestaBoard transitions, Vestaboard transition, flip animation, transition plugins, typewriter transition, dissolve transition, split-flap animation, Transition Lab]
+---
+
+# Transitions
+
+Transitions control how your board gets from the message it is showing now to the next one — a left-to-right wave, a dissolve, a typewriter reveal, or nothing at all.
+
+## Overview
+
+FiestaBoard has **two different kinds of transitions**, and they behave differently. Knowing which one you are looking at explains most of the surprises people run into.
+
+| | Built-in flip strategies | Transition plugins (beta) |
+|---|---|---|
+| **Who animates** | The Vestaboard itself | FiestaBoard, by sending many frames |
+| **Works on** | Local API connections only | Any board connection (Local or Cloud) |
+| **Examples** | Wave, Drift, Curtain, Row, Diagonal, Random | Typewriter, Simple Dissolve, Slot Machine, Quiet Library |
+| **Where to turn on** | Always available | **Settings → Advanced → Beta Features → Transition Plugins** |
+| **Step Interval / Step Size** | Supported | Not used — plugins set their own pacing |
+| **Stability** | Stable | Experimental; behavior and settings may change |
+
+## Built-in Flip Strategies
+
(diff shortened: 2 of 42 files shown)


jeffredodd:
## Production verification

Verified against a **production image** (`--target runtime`) on an isolated network with the mock board — not the dev container, not jsdom.

### A real bug the unit tests could not catch

The first e2e run failed on the per-page picker:

```
<div role="presentation" data-base-ui-inert=""> from <div data-base-ui-portal=""> subtree intercepts pointer events
119 × waiting for element to be visible, enabled and stable
  > await page.getByRole("button", { name: /save/i }).first().click();
```

Base UI's `MenuRadioItem` defaults `closeOnClick` to **false**. Picking a transition left the menu open, and its inert backdrop swallowed the Save click — the setting was selectable but not saveable without dismissing the menu by hand.

jsdom can't see this: Testing Library dispatches clicks straight at the node and never consults the backdrop's `pointer-events`. All 13 unit tests passed against the broken build. Fixed in 8f884aa; the same test now completes in 2.4s instead of timing out.

This is the app's first `DropdownMenuRadioItem`, so there was no local convention to inherit the right prop from.

### The transition actually drives the board

Frame counts from the mock board — persisting a setting is not the same as it working:

| Surface | Setting | Frames delivered |
|---|---|---|
| Global | none | 1 |
| Global | `plugin:typewriter` | **133** |
| Per-page (global = none) | none | 1 |
| Per-page (global = none) | `plugin:slot_machine` | **28** |

The per-page rows are the important ones: the global default was `null` throughout, so those 28 frames came from the page override alone.

### API behaviour in prod

```
GET /api/transitions/plugins        -> 404   (beta off)
PUT /api/settings/beta              -> 200
GET /api/transitions/plugins        -> ['quiet_library','simple_dissolve','slot_machine','typewriter']
GET /api/plugins                    -> typewriter=transition, date_time=data, ...
```

### E2E

`web/tests/transition-pickers.spec.ts`, 4 tests, green against the production build:

```
✓ global picker saves a transition plugin as plugin:<id> (3.2s)
✓ global picker hides plugin options when the beta is off (1.3s)
✓ page picker saves an override and clears it back to the global default (2.6s)
✓ integrations badges a transition plugin and offers no enable toggle (323ms)
4 passed
```

Non-vacuity, since two of these passed before the fix and needed proving:
- **Global picker** — re-ran with the strategy pre-seeded to `diagonal`. It still ended at `plugin:typewriter`, so the click causes the change rather than the assertion matching a coincidence.
- **Integrations** — carries a permanent positive control: a data plugin in the same table must still render its toggle and must not carry the badge. Without it, `toHaveCount(0)` would also pass if the selector simply stopped matching anything.
- **Page picker** — proven by the failure above; it is the test that caught the bug.

Reproduce:

```bash
docker build --target runtime -t fb-transitions:prod .
docker network create fbtp-net
docker run -d --name fbtp-mock --network fbtp-net \
  -v "$PWD/integration-tests/mock-board/server.py":/server.py:ro \
  -e PORTS=7000,7001 python:3.11-alpine python /server.py
docker run -d --name fb-tp-verify --network fbtp-net -p 4499:3000 \
  -e PRODUCTION=true -e FIESTABOARD_AUTH_ENABLED=false fb-transitions:prod
docker run --rm --network fbtp-net -v "$PWD/web":/app -w /app \
  -e BASE_URL=http://fb-tp-verify:3000 -e MOCK_BOARD_HOST=fbtp-mock \
  -e MOCK_BOARD_URL=http://fbtp-mock:7000 \
  mcr.microsoft.com/playwright:v1.62.1-noble \
  npx playwright test tests/transition-pickers.spec.ts
```

PR #1597, opened 2026-08-13
refactor(board): migrate to @fiestaboard/ui's BoardDisplay, and mark Standard as the default flip speed

### LINKED ISSUE #1550: Expose board transition speed as a user setting (FiestaUI flapSpeed)
FiestaUI now exposes split-flap transition speed as a prop, so this can become a real user setting.

## What shipped upstream

`@fiestaboard/ui` **v2.1.1** (Fiestaboard/FiestaUI#189, closing FiestaUI#178) added a `flapSpeed` prop to `BoardDisplay`:

```ts
flapSpeed?: "hardware" | "quick" | "standard" | "relaxed" | { durationMs: number }

FLAP_SPEED_PRESETS = {
  hardware: 16,   // the device's own rate: 60 RPM x 62 flaps = 16.1ms/flap
  quick:    48,
  standard: 80,   // default — unchanged from every previous version
  relaxed: 130,
}
```

`standard` is the default, so nothing changes until a user opts in. The escape hatch is a tagged object (`{ durationMs }`) so it can never be confused with a preset, and the value is clamped to [8, 2000].

That one number drives the CSS leaf animations, the cast shadow **and** both `setInterval` cadences — there is no second constant to keep in sync (`LOADING_TICK_MS` was deleted in the same change).

## Why `hardware` is worth offering but not defaulting

The 80ms default is a deliberate ~5x deviation from the real device. At the hardware's own ~16ms a Latin glyph is on screen for roughly one frame at 60Hz and is not resolvable mid-cascade — a settling board reads as grey noise rather than letters turning over. Some users will want true-to-device anyway, which is exactly why it is a preset rather than a hardcoded choice.

Only `hardware` claims a device derivation. `quick` / `relaxed` are deliberately generic names rather than invented Vestaboard mode names.

## ⚠️ Blocked on an upstream bug

**Do not build this yet.** Fiestaboard/FiestaUI#196 — the board currently snaps to a new message instead of cascading, so `flapSpeed` has nothing to control on a message change. Measured: zero flap animations run, and all four presets settle identically.

That needs fixing first, otherwise the setting will appear to do nothing.

## Suggested shape once unblocked

- Settings entry with the four presets, `standard` selected by default and labelled as such.
- Live preview so the choice is evaluable before saving — the difference between `quick` and `standard` is not obvious from a label.
- Persist per user, not per board, unless there is a reason to vary it per board.
- Note in the copy that `hardware` matches the physical unit and is intentionally hard to read mid-flip.

## Naming

The upstream preset words are settled (`hardware` / `quick` / `standard` / `relaxed`), so the UI labels can track them directly rather than inventing a parallel vocabulary.

jeffredodd:
Closes #1550.

**Read this first: #1550's premise has moved on since it was written.** The setting itself already shipped — #1555 landed the cascade fix, the four presets, the live preview, the reduced-motion snap and all 14 locales. What was still outstanding was the decision that PR deliberately left open: **(a) migrate the app onto `@fiestaboard/ui`'s `BoardDisplay`, or (b) keep porting fixes into the app's fork.** The call was (a). This is (a), plus the one product requirement in #1550 that was genuinely unmet.

---

## 1. The parity assessment, done before any code

The app rendered a ~1,400-line fork of `@fiestaboard/ui`'s `BoardDisplay`, plus forks of `StaticBoardDisplay` (402 lines) and `ScaledBoardDisplay` (243 lines). I read all three against the package's `v2.2.6` sources in full — component, siblings (`board-surfaces`, `reduced-motion`, `board-metrics`) and libs (`board-characters`, `board-colors`, `board-dimensions`) — before touching anything.

### Behaviour → covered by the package?

| Behaviour the app depends on | Covered? | Evidence |
|---|---|---|
| `BOARD_CHARS` 0–71, placeholders at 43/45/51/57/58/61, colour codes `"63"`–`"71"` | **Identical** | `board-characters.ts:24–101`, element-for-element |
| `EXTRA_CHARS` ♥, `getCharIndex`, `isColorTile`, `getCharFromToken`, `tokensEqual` | **Identical results** | `board-characters.ts:106–158` (Map/Set instead of `indexOf`/`includes`) |
| ° → ♥ substitution on `deviceType="note"` | **Identical**, and reaches further | `applyDeviceSubstitution`, used by `messageToGrid` *and* the accessible name (app's fork announced "°" while drawing ♥) |
| `{63}` / `{red}` / `{/red}` / `{/}` parsing, invalid `{…}` falling through as literals, uppercasing | **Same algorithm** | `board-characters.ts:167–211` |
| **White-board inversion** (69↔70 swap on `boardType="white"`) | **Identical, character-for-character** | `board-colors.ts:114–120` — same function, same doc comment, same hexes, same fallback; applied at all four call sites |
| Palette hexes | **Identical** | `board-colors.ts:26–35` |
| **Note-array seams** `data-note-row` / `data-note-tile` / `data-note-row-seam` / `data-note-col-seam`, seamGap 6/8/10px, boundaries at `% NOTE_ROWS` / `% NOTE_COLS` | **Identical** — same attributes, same values, same elements | `board-display.tsx:463–482, 527–545, 1311–1313` |
| **Draw-mode metadata** `emitCellMetadata` → `data-row` / `data-col` / `data-cell-value`, off by default, `data-cell-value` on the static path only | **Identical, including the asymmetry** | same element (the wrapper), same values, same default |
| Tile test hooks `char-tile-{r}-{c}`, `data-current-char`, `data-target-char`, `data-is-animating`, `data-is-transitioning` | **Identical** — same names, same element, same sources | `board-display.tsx:995–999` |
| `data-board-preview`, `role="img"`, `aria-hidden` inner bezel | **Identical** | the `globals.css` animation kill switches key on this and are unaffected |
| **`prefers-reduced-motion` snap** | **Covered and hardened** | package ANDs its own `useReducedMotion()` (`useSyncExternalStore`, SSR-safe) on top of the prop, and the loading ticker refuses to start under `reduce`. The setting *cannot* buy past it |
| `isStatic` fast path | **Same node count and nesting** | `board-display.tsx:381–490` |
| `ScaledBoardDisplay` fit/actual toggle | **Same `sessionStorage` key** (`fiestaboard:boardPreviewMode`) **and same `data-testid="actual-size-scroll"`** | it was extracted from this app |
| `animationsEnabled` source | **Wiring** — internal hook → prop | supplied by `useBoardChrome` |
| `flapSpeed` default | **Wiring** — package defaults to `"standard"`, app defaults to the user's setting | supplied by `useBoardChrome` |
| i18n'd accessible names | **Wiring** — `loadingLabel` / `emptyLabel` / `messageLabel` props | supplied by `useBoardChrome` / the static wrapper |
| `announceUpdates` polite live region | **Package only**, opt-in, default off | not wired here — see follow-ups |

**Nothing the app depends on is missing from the package.** Every gap is wiring-level and fixable here, which is why this is a migration PR and not an issue comment.

### Divergences that are behaviour changes

| | App fork | Package | Effect |
|---|---|---|---|
| `ALL_COLOR_CODES` prototype | plain object — `{toString}` parsed as a colour, rendered a broken tile | `Object.create(null)` | `{toString}` now renders as literal text. Adversarial input only |
| `resolveDimensions` | unclamped | clamps to `[1, 8]` notes/axis, caches for stable identity | the app never renders past 8 (`isValidNoteArrayGrid` enforces it), so inert |
| Accessible name derivation | regex-stripped raw message | parsed with the same parser the tiles use: uppercased, ♥-substituted, whitespace-collapsed; colour-only board named `"Board display"` instead of a dangling `"Board display: "` | strictly better; see follow-up 3 |
| `data-slot="box"` / `"text"` on every tile node | present (app used `Box`/`Text` primitives) | raw `div`/`span`; `data-slot="board-display"` on the root | nothing in this repo selects those inside a board — checked |
| keyframes injection | imperative `document.head` mutation **during render** | `<style href="board-flap-keyframes" precedence="default">`, React-19-hoisted, only on the animated path | render-safe / React-Compiler-safe |

---

## 2. The visual delta — this is the part that needs eyes

The package's board carries four upstream changes the fork never got. **No functionality is lost, but the board looks different on every surface.** Precisely:

**FiestaUI #176 — tile geometry derived from one ratio set** (the fork had four independent class maps that had drifted; `md`'s gap had silently lost its `lg:` step):

| size / step | tile (before → after) | glyph | gutter | corner radius |
|---|---|---|---|---|
| `sm` | `14×18` → **`13×18`** | 7px | 3px | 3px → **1.35px** |
| `md` base | `14×20` | 7 → **8px** | 2 → **3px** | 3px → **1.5px** |
| `md` `sm:` | `20×28` | 10 → **11px** | 4px | 3px → **2.1px** |
| `md` `md:` | `24×34` | 13px | 5px | 3px → **2.55px** |
| `md` `lg:` | `28×40` | 16px | **5 → 6px** (the missing step) | 3px |
| `lg` base | `18×26` | 10px | 3 → **4px** | 3px → **1.95px** |
| `lg` `lg:` | `32×46` | 20 → **18px** | 7px | 3px → **3.45px** |

A 22-column flagship: `sm` **371 → 349px** (~6% narrower); `md` on a phone **350 → 371px** and ~5px taller; `md` at ≥1024px **721 → 742px**; `lg` desktop unchanged in width, glyph one step smaller. Bezel padding and border/rounding are **identical**.

**FiestaUI #177 — single-leaf flip handoff.** Before: top leaf `[0, 0.55·step)`, bottom `[0.35·step, 0.9·step)` — both leaves in motion for 20% of every step (a tile visibly splitting open at the midpoint, which a physical module cannot do) and 10% dead time. After: top `[0, step/2)`, bottom `[step/2, step)`, tiling the step exactly. Easings and the cast shadow are unchanged.

**FiestaUI #179 — leaf materials.** The hinge seam goes from a 1px flat `rgba(0,0,0,0.3)` line (≈4/255 luminance delta on a black board — effectively invisible) to a 2px gap-plus-highlight gradient, and character tiles gain a leaf bevel. **On the default black board this is the most noticeable change: the board reads as flaps rather than as a grid of dark rectangles.**

**FiestaUI #200 — no `max-w-full` on the bezel.** The fork clamped the frame below the grid's min-content width, so in a narrow slot the rows ran past both bezel edges. The package overflows honestly instead and expects `ScaledBoardDisplay` to do the fitting. Every board surface in this app already goes through `ScaledBoardDisplay` except the settings live preview (a `sm` Note ≈ 237px, comfortably inside a settings card) and Storybook.

**I could not run this in a browser** — the machine is running several other agents' containers and a full dev-container build was not a responsible use of it. The deltas above are read off the sources, not guessed, but *someone should look at a board before this merges.* That is the one check I am not claiming.

---

## 3. What changed

`board-display.tsx` (1,448 → 82 lines), `static-board-display.tsx` (416 → 32), `scaled-board-display.tsx` (254 → 36). Net **−1,759 lines**.

`useBoardChrome(flapSpeed?)` is the single place the app's four concerns are supplied, shared by the plain and scaled wrappers so the two cannot drift:

- `animationsEnabled` — still `useBoardAnimationsEnabled()`, so `board_animations` (`on`/`desktop`/`off`), `reduce_motion` and `prefers-reduced-motion` behave exactly as before. The package ANDs its own reduced-motion check on top, so this is belt-and-braces.
- `flapSpeed` — still the user's `board_flap_speed`; an explicit prop still wins (only the settings live preview passes one).
- `loadingLabel` / `emptyLabel` / `messageLabel` — `messageLabel` goes through `useCallback` because the package's memo comparator compares it **by reference**, and an inline arrow would defeat memoization of a ~132-tile grid on every parent render.

`StaticBoardDisplay` keeps passing `previewLabel` (a fixed name) rather than letting the package derive a name per thumbnail — preserving what the page grid announces today. The derived name is better; changing it does not belong in a migration.

Second commit: `standard` is now labelled **"Standard (default)"** on the option itself. #1550 asked for it "labelled as the default"; today the only thing saying so is the hint under the group, which describes whichever preset is *selected* — so a user sitting on Relaxed has no way to tell what they moved away from. One combined string per locale (`flapSpeed_standardWithDefault`) rather than an appended "(default)", so translators own the whole label. All 14 locales, additive, one line each.

---

## 4. Non-vacuity evidence

Per CLAUDE.md, a test that still passes when you revert the code it covers is testing nothing.

**`board-display-labels.test.tsx` is the suite that needed proving.** The package's English label defaults are *byte-identical* to this app's `en.json` (`"Loading board display"`, `"Empty board display"`, `` `Board display: ${msg}` ``), so an English assertion would pass whether or not the wrapper wires anything up. The file therefore replaces the global English i18n mock with a sentinel `t` no fallback can imitate.

Observed, with the three label props removed from `useBoardChrome` and from the static wrapper and nothing else changed:

```
❯ src/__tests__/board-display-labels.test.tsx (5 tests | 5 failed)
  × names a board showing a message with the app's withMessage translation
      AssertionError: expected 'Board display: HELLO' to contain 'boardDisplay.withMessage'
  × names an empty board with the app's empty translation
      Expected the element to have attribute: aria-label="[[boardDisplay.empty]]"
      Received:                               aria-label="Empty board display"
  × names a loading board with the app's loading translation
      Expected the element to have attribute: aria-label="[[boardDisplay.loading]]"
      Received:                               aria-label="Loading board display"
  × names a scaled board with the app's translations too
      AssertionError: expected 'Board display: HELLO' to contain 'boardDisplay.withMessage'
  × names a static preview with the app's preview translation
      Expected the element to have attribute: aria-label="[[boardDisplay.preview]]"
      Received:                               aria-label="Board preview: HELLO"

Test Files  1 failed (1)
     Tests  5 failed (5)
```

Restored → 5 passed. Every failure is the package's English default leaking through, which is exactly the regression the suite exists to catch.

**The default-marker tests were written first and watched fail** before the locale key or the component change existed:

```
× offers the four presets with the stored one selected
    AssertionError: expected [ 'Hardware', 'Quick', …(2) ] to deeply equal [ 'Hardware', 'Quick', …(2) ]
× marks Standard as the default even when another preset is selected
    TestingLibraryElementError: Unable to find role="radio" and name "Standard (default)"
× falls back to Standard when the stored value is a raw millisecond count
    TestingLibraryElementError: Unable to find role="radio" and name "Standard (default)"

Test Files  1 failed (1)
     Tests  3 failed | 4 passed (7)
```

**The 55 pre-existing board tests are the migration's real contract** and were not touched. They pass unchanged, which is the load-bearing result here — in particular `snaps with no flap layers when the board-animations kill switch is off`, `snaps with no flap layers under prefers-reduced-motion, whatever the speed`, `settles in a time that scales with the chosen flap speed`, `retargets in place when a second message lands mid-cascade`, the white-board colour inversion, the ° / ♥ substitution, the note-array seams, the draw-mode `data-cell-value` gating, and the scaled fit/actual toggle.

---

## 5. Verification

All in `node:24-alpine` against the same mounted `web/`, per CLAUDE.md (Docker only). The baseline is `origin/main` extracted to a clean tree and run through the identical commands.

| | `origin/main` | this branch |
|---|---|---|
| `vitest run` | `Test Files 121 passed (121)` · `Tests 1499 passed \| 13 skipped (1512)` · 1 unhandled error (pre-existing TipTap `window is not defined` in `page-builder-device-switch`) · exit 1 | **`Test Files 122 passed (122)` · `Tests 1505 passed \| 13 skipped (1518)` · 0 errors · exit 0** |
| `npm run lint` | `378 problems (0 errors, 378 warnings)` · exit 0 | **identical: `378 problems (0 errors, 378 warnings)` · exit 0** |
| `npm run format:check` | `All matched files use Prettier code style!` · exit 0 | **same · exit 0** |
| `npm run typecheck` | 136 errors · exit 2 (pre-existing, being fixed in parallel) | **136 errors · exit 2 — same count, and the 6 in `board-display-transition.test.tsx` are byte-identical on both sides. Zero errors added.** |

Delta is exactly `+1 file / +6 tests` — the 5 label tests and the default-marker test.

One earlier run of the branch suite reported `120 passed (122)` with 2 unhandled errors; that was worker OOM from other agents' containers competing for the machine (an `origin/main` run crashed the same way at `81 passed (121)`). Re-run with `--maxWorkers=2` it is clean, as above. Flagging it because the parenthesised total is the number that matters and it is easy to misread.

**Not run, and not claimed:** the Playwright suites (`e2e-tests`, `ai-mcp-e2e`, `auth-e2e`, `ha-ingress-e2e`) need a Docker Hub login plus mock board/cloud/LLM containers; the Docker image build needs the same. The Storybook a11y job was not run either. **These are the checks most exposed by this PR** — a board that now renders a different DOM and different geometry. `mobile-critical-flows.spec.ts:102` (document `scrollWidth` ≤ viewport + 20px) is the specific one to watch, given #200 removed the bezel's `max-w-full`; every mobile board surface goes through `ScaledBoardDisplay`, which is exactly what that change assumes, but I have not proven it.

---

## 6. What this covers, and what it does not

**Covered.** The migration, with every app behaviour preserved (verified above). `flapSpeed` reaching the package's board as the user's chosen cadence. `prefers-reduced-motion: reduce` still snapping with no flap layers regardless of preset. `standard` labelled as the default.

**Not covered, deliberately:**

1. **A human looking at a board.** Section 2 is exact but it is read from source. This should not merge on my say-so.
2. **`announceUpdates`.** The package can announce message changes through a polite live region (FiestaUI #206). The dashboard's live board is precisely the surface it was built for — a mirrored board where a change is news — but it is a new behaviour with no coverage here, so it stays off. Worth a follow-up.
3. **A genuine upstream i18n gap.** `NO_TEXT_LABEL = "Board display"` is hardcoded English in the package with no prop to override, and it is what a colour-only board announces. The app is trading a dangling `"Board display: "` for an untranslatable string on that one edge case — a net improvement in clarity, a small regression in localizability. The clean fix is a `noTextLabel?: string` prop upstream; I did not open that PR because it does not block this one and needs its own release cycle. Worth filing on FiestaUI.
4. **`StaticBoardDisplay`'s per-thumbnail name.** Still a fixed `previewLabel`, preserving today's behaviour. The package's default (name each preview by its own text) is better a11y and is a one-line change once someone decides the page grid should announce that way.
5. **Persistence shape.** `board_flap_speed` lives on `DisplaySettings` (server-side `settings.json`, per install), alongside `reduce_motion` / `board_animations` / `site_animations`, where #1555 put it. Theme persists in `localStorage` instead, so "per user" is arguably the other convention — but moving it would be a migration for no user-visible gain, so I left it. Flagging it rather than churning it.


jeffredodd:
## Mobile clipping fixed — `6a6a9079`

Reproduced, diagnosed, fixed and re-measured against real builds on a 390×844 viewport. Thanks for catching it; the root cause was not what either of us first assumed.

### It is not the missing `max-w-full` — measured, not reasoned

I injected `max-width: 100%` on the bezel, as you did, and measured one level deeper:

| | container | bezel | leftmost tile |
|---|---|---|---|
| as shipped | 30 → 345 (315px) | **-2 → 377 (379px)** | **x=13** |
| with `max-w-full` restored | 30 → 345 (315px) | 30 → 345 (315px) ✅ | **x=13** ❌ |

The bezel snaps flush, and the first tile does not move. The row is `justify-center` over `shrink-0` tiles, so 349px of tiles centre themselves in a 285px line box and escape ~32px on **each** side. `max-w-full` never made this board fit — it only stopped the frame from advertising the overflow.

### The actual root cause

This surface never had scaling, and never needed it, because the app's fork could **distort**. Its static tiles had no `shrink-0`, so a flex row squeezed them toward the glyph width and the board "fitted" by silently breaking its own proportions. FiestaUI #203 added `shrink-0` and #200 dropped `max-w-full`, on an explicit principle stated in the package source:

> With `shrink-0` a squeezed board overflows instead of distorting, which is both honest and what ScaledBoardDisplay's measurement of the board's natural width assumes. Consumers with a slot too narrow for a board want ScaledBoardDisplay (issue #192).

So the migration traded a *distorted* board for a *clipped* one. Both were wrong; the new one is visibly wrong. **The fix belongs in this app, not upstream** — restoring `max-w-full` in FiestaUI would re-break the invariant #200 fixed and, per the measurement above, would not fix the clipping anyway. No FiestaUI PR needed.

### The fix

`PageButtonPreview` now renders `ScaledBoardDisplay` with `isStatic`. `isStatic` keeps the zero-hooks-per-tile path that made this a `StaticBoardDisplay` in the first place — the added cost is per board, not per tile, and the component already defers mounting until in-viewport. The right-fade mask and the `-mr-3` bleed are gone; they existed only to disguise a board that never fitted.

Measured after, same viewport:

| | before | after |
|---|---|---|
| transform | **none** (all 8 ancestors) | **`scale(0.831)`** |
| tile span | 349px | **290px** |
| container | 315px | 315px |
| overflow L / R | **+17 / +17** | **-12 / -12** |

At **1440px the transform is `scale(1)`** and the board is pixel-identical to before — `ScaledBoardDisplay` never scales up, so desktop is untouched.

### Audit of every preview surface

| Surface | Component | Verdict |
|---|---|---|
| Dashboard Active Display | `ScaledBoardDisplay` | ✅ already correct — measured 269px in 285px |
| Page builder | `ScaledBoardDisplay` | ✅ |
| Wizard board setup | `ScaledBoardDisplay` | ✅ |
| Inline board preview (AI chat) | `ScaledBoardDisplay` | ✅ |
| **Pages list card** | ~~`StaticBoardDisplay`~~ → `ScaledBoardDisplay isStatic` | ❌ **fixed** |
| Collection cascade | `StaticBoardDisplay` | ✅ **left as-is, deliberately — measured** |
| Settings flip-speed preview | raw `BoardDisplay`, `sm` Note | ✅ ~237px, fits |

On the cascade: I measured it rather than assuming. Its cards are absolutely positioned and start **inside** the 160px frame — `overflowLeft` of **-15px** and **-33px** — bleeding off the *right* by design. That overlap is the cascade. Nothing is cut off at the start of a row, which is the defect scaling exists to prevent. Wrapping it would also measure a slot with no definite width and add the Fit/Actual toggle to a thumbnail for note-array pages.

### The regression test

`web/tests/regression/board-preview-fit.spec.ts` — 5 tests, geometry not presence, sweeping **every** board on the page rather than naming a component, because this recurred once already (#1397) and the next one lands wherever nobody enumerated.

Two invariants, deliberately not merged: overflowing **right** can be design (the cascade); overflowing **left** never is, because it cuts the first character off every row. So the sweep asserts the left edge everywhere, and the surfaces meant to show a whole board get the stronger "fits entirely" check.

**Failing first, against the unfixed build (`fiestaboard:pr1597` at `493967c7`):**

```
✘ no board on the Pages list is cut off at its left edge
✓ no board on the dashboard is cut off at its left edge
✘ the Pages list previews fit their container entirely at 390px
✓ the dashboard board preview fits its container entirely at 390px
✓ no board pushes the document wider than the viewport at 390px

  1) no board on the Pages list is cut off at its left edge
     Error: boards cut off at the left edge: [
       { "slot": "static-board-display",
         "clipClass": "w-full hover-stable overflow-hidden -mr-3",
         "clipWidth": 315, "tileSpan": 349,
         "overflowLeft": 17, "overflowRight": 17 },  … ×4
     ]
```

It fails on the broken surface and passes on the three already-correct ones — it discriminates, rather than going red wholesale.

**Passing, against the fixed build:** `5 passed (1.8s)`.

**One vacuity trap I had to design out.** My first draft filtered on `slot === "static-board-display"`. The fix *changes which component renders the board*, so that filter would have matched nothing the moment it started working and passed forever. It now filters on geometry only and asserts `boards.length > 0` first, with the message `"measured no boards — the check would pass vacuously"`.

### Verification

| | baseline (`origin/main`) | this branch |
|---|---|---|
| vitest | 121 files (121) · 1499 tests | **122 files (122) · 1505 tests · exit 0** |
| lint | 378 problems (0 errors) | **378 problems (0 errors)** |
| format:check | clean | **clean** |
| typecheck | 136 errors | **136 — none added** |
| Playwright (`board-preview-fit` + `dashboard` + `note-arrays`) | — | **27 passed** |

PR #1661, opened 2026-08-21
chore(deps): bump the npm-dependencies group in /web

jeffredodd:
Reconstitutes the 15-package group update from #1655.

Dependabot auto-closed that PR — *"Looks like these dependencies are updatable in another way, so this is no longer needed"* — when #1659 landed the `nanoid` override and changed how the group resolved. The updates themselves were never applied, so this re-applies them.

## Direct dependencies

| Package | From | To |
|---|---|---|
| `@tiptap/extension-paragraph` | 3.29.2 | 3.30.1 |
| `@tiptap/pm` | 3.29.2 | 3.30.1 |
| `@tiptap/react` | 3.29.2 | 3.30.1 |
| `@tiptap/starter-kit` | 3.29.2 | 3.30.1 |
| `lucide-react` | 1.30.0 | 1.31.0 |
| `sonner` | 2.0.7 | 2.0.8 |
| `@storybook/addon-a11y` | 10.5.7 | 10.5.8 |
| `@storybook/react-vite` | 10.5.7 | 10.5.8 |
| `@testing-library/jest-dom` | 7.0.0 | 7.0.1 |
| `@testing-library/user-event` | 14.6.3 | 14.6.4 |
| `rollup-plugin-visualizer` | 7.0.1 | 7.1.1 |
| `storybook` | 10.1.11 | 10.5.8 |
| `typescript-eslint` | 8.66.0 | 8.67.0 |

Plus the two transitive platform binaries dependabot listed — `@rolldown/binding-linux-x64-gnu` and `@swc/core-linux-x64-gnu` — which come along via the lockfile.

All minor/patch, no majors. A few resolve slightly ahead of dependabot's targets (tiptap 3.30.2, lucide-react 1.33.0, storybook 10.5.10) because more releases landed in the interim; all sit inside the same `^` ranges.

The `nanoid` 3.3.18 override from #1659 is preserved — verified in the regenerated lockfile.

## Verification

Lockfile regenerated with `npm install --package-lock-only --legacy-peer-deps`, then a clean `npm ci` on node:24 in a container:

```
format:check  All matched files use Prettier code style!   exit=0
lint                                                       exit=0
typecheck                                                  exit=0

Test Files  125 passed (125)
     Tests  1474 passed | 7 skipped (1481)
  Duration  39.00s
```

`format:check` is called out explicitly because `Lint UI (web)` fails on `prettier --check`, not just eslint.

Co-Authored-By: Claude <noreply@anthropic.com>


jeffredodd:
Superseded by #1660. Dependabot did recreate the group after all — #1660 is a strict superset of this PR (the same 15 packages plus `@codemirror/commands` 6.10.4 -> 6.11.0 and `@codemirror/view` 6.43.8 -> 6.43.9), and it is based on main after #1659 so it keeps the `nanoid: ^3.3.18` override.

Closing this in favour of the dependabot-owned PR so the group stays under dependabot's management rather than drifting from a hand-rolled copy.

PR #1664, opened 2026-08-21
fix(board): let each board say which glyph its code-62 flap carries

### LINKED ISSUE #1657: Bug: code 62 renders as ° on Flagship, but current hardware ships the heart
# Bug: code 62 renders as `°` on Flagship, but current hardware ships the heart

## Summary

FiestaBoard picks the glyph for character code 62 from `device_type` — `°` on
Flagship, `❤` on Note. Vestaboard has since
[replaced the degree symbol with a heart on the Flagship](https://www.vestaboard.com/news/heart-character-comes-to-vestaboard),
so current Flagship units render 62 as a heart and the preview no longer matches
the board.

## Environment

- FiestaBoard 8.27.30, Docker install
- Vestaboard Flagship (white), configured as `device_type: flagship`,
  `board_color: white`

## Steps to reproduce

1. Configure a current-production Flagship board.
2. Create a page containing character code 62 (e.g. a temperature like `72°`).
3. Compare the in-app preview to the physical board.

**Expected:** preview matches the board.
**Actual:** preview shows `72°`; the board shows `72❤`.

## What I found

I traced this with Claude Code, so the references below are more thorough than my
own familiarity with the codebase would account for. I've read and checked them,
but flagging it — it's your codebase, and you'll spot anything I've misread
faster than I will.

The Flagship-means-degree assumption appears in:

| Location | What it does |
|---|---|
| `src/board_html_renderer.py:114` (and `:273`) | `is_note = device_type == "note"` gates the `°` → `♥` substitution in `_grid()` |
| `web/src/components/tiptap-template-editor/components/ColorPickerContent.tsx:35,200` | The heart button in the character palette is gated on `isNote`, so on a Flagship it can't be inserted from the picker |
| `@fiestaboard/ui` → `src/lib/board-characters.ts:220` | `applyDeviceSubstitution()`, called from `messageToGrid` (`:238`) and `messageToText` (`:285`) |
| `src/board_chars.py:81,97`, `docs/reference/character-codes.md:118-120` | State the rule as documentation |

One thing I wasn't sure whether to mention: `messageToText` also builds the
board's accessible name, so on my board a screen reader announces "degree" for a
tile that's drawing a heart. The comment at `board-characters.ts:218` suggests
tiles and announcements agreeing was deliberate, so it might be worth keeping in
scope for whatever the fix turns out to be.

`web/src/__tests__/board-display-colors.test.tsx:99` asserts that a Flagship
renders `°`, so this looks like a deliberate contract rather than an oversight —
which is why I'm filing rather than sending a patch.

## Possible approach

The closest existing pattern I could find is `board_color` (`src/devices.py:106`)
— a per-board attribute set during the wizard, editable later in Settings →
Hardware, and passed down to the display. A similar per-board field for code 62,
defaulting to current behaviour, would leave existing installs untouched.

Noting it as a starting point rather than a recommendation — you'll know whether
it fits the architecture better than I do.

## The FiestaUI dependency

Whichever shape the app-side setting takes, the preview substitution lives in the
package: `applyDeviceSubstitution` (`src/lib/board-characters.ts:220`) takes an
`isNote` boolean, shared by `messageToGrid` and `messageToText`. Since both call
sites already route through that one function, the package change may be
contained — but it's a FiestaUI release either way, which I can't do.

---

If it's useful, I have a current-production Flagship on hand and am glad to test
a fix against real hardware.

jeffredodd:
Closes #1657

> [!IMPORTANT]
> **Draft: this blocks on a `@fiestaboard/ui` release.**
> The in-app preview substitution is package-owned, and the fix for it is Fiestaboard/FiestaUI#247, which is open and unreleased. Once the maintainer cuts that release, `web/package.json` needs bumping off `5.1.0` and the lockfile updating; then this is ready. No version bump is faked here — the pin is untouched, so **CI will fail typecheck on `code62Glyph` until the bump lands.** Everything below was verified locally against the package built from that branch.

## The problem

Character code 62 is **one code with two possible physical flaps**. Vestaboard replaced the Flagship's degree flap with a heart on newly-manufactured units:

> Every new Vestaboard purchased will ship with the heart in place of the degree symbol, and the first displays with the heart character are on their way to customers now.

They published no serial or date boundary. So two boards that both report `device_type: "flagship"` draw different glyphs, **nothing FiestaBoard can query tells them apart**, and the reporter's Flagship (ordered Dec 2025, shipped Aug 2026) shows `72♥` where every FiestaBoard preview shows `72°`.

## What this adds

`BoardInstance.code62_glyph` — `"degree" | "heart"`, per board, in the shape of `board_color` (`src/devices.py`). Set it in the setup wizard when adding a board, or later in **Settings → Hardware**.

**Default is `"degree"`** — the glyph every Flagship carried before the change — so no existing install renders differently until its owner says otherwise.

**Flagship only.** `effective_code62_glyph` is what render paths read, and Note / note-array hardware always resolves to `"heart"`: those flaps only ever carried the heart, so the setting is not theirs to answer, and a board switched Flagship→Note must not draw a degree sign it does not physically have from a leftover preference. The control is not rendered for Note boards.

**Display-only.** Both glyphs are character code 62 on the wire. `src/board_chars.py` still maps all three of `°`, `❤` and `♥` to 62 and no second character code is introduced — there is a test asserting exactly that.

## No migration, and why

`BoardInstance.from_dict` reads every key with a default, and `to_dict` is `asdict`. A board stored before this field existed loads as `"degree"` — which is precisely what it already drew — and round-trips the new value once set. **The stored format gains a key but no stored board changes meaning**, so an ordered migration would have had nothing to change and none was added. That claim has its own test (`test_a_board_stored_before_this_field_existed_loads_as_degree`), so if `from_dict`'s default is ever removed the suite says so rather than the field silently vanishing on every save.

## One rule, four places

`resolveCode62Glyph` is mirrored in `web/src/hooks/use-board.ts`, `_resolve_code62_glyph` in `src/board_html_renderer.py`, `effective_code62_glyph` in `src/devices.py`, and (upstream) `resolveCode62Glyph` in `@fiestaboard/ui`. The dashboard, the editor preview, and the MCP HTML preview therefore cannot disagree about the same board. Each mirror has a test pinning the Note case.

## Also fixed: the picker had no way to insert code 62 on a Flagship

`ColorPickerContent`'s heart button was gated on `deviceType === "note"`, so a Flagship owner could not reach character code 62 from the picker at all — including the owners of the 2026 boards this issue is about. It is now offered for every board, and only its **wording** follows the board: a degree-flap Flagship gets a `°` button labelled "Degree", not a button captioned "Heart" that draws a degree sign. It inserts `°` in every case, because that is what goes on the wire.

## Accessible names

The reporter raised this specifically. `messageToText` builds the `role="img"` name from the same substitution as the tiles, so a board drawing ♥ announces ♥ — a name saying "degree" for a heart tile would be a text alternative for a different image (WCAG 1.1.1). That property is enforced upstream in FiestaUI#247, including a sweep over every device/preference pair and a test that flips the setting on a mounted board and requires the name to follow.

## Docs

`docs/reference/character-codes.md` stated Flagship=degree / Note=heart as fact. Rewritten for users: a three-row table (Note, pre-2026 Flagship, 2026+ Flagship), where the setting lives, that the default is Degree so older-board owners need do nothing, that it is per board so a mixed household previews each correctly, and that it never changes what is sent. `src/board_chars.py`'s `DEGREE`/`HEART` comments carried the same stale rule and now explain the two-flap reality.

## i18n

Nine new keys across all **14** locale files (`code62Label`, `code62Help`, `code62DegreeAriaLabel`, `code62HeartAriaLabel` in both `displaySettings` and `wizard.boardSetup`; `degreeLabel`, `degreeCharacterAriaLabel`, `insertDegreeTooltip` in `templateEditor`), plus a correction: `insertHeartTooltip` said "(Note only)", which is no longer true. `i18n-config.test.ts` enforces exact key parity across all 14 and passes.

## Test evidence

Green:

- **Python** — 450 pass across `test_devices.py`, `test_board_html_renderer.py`, `test_board_chars.py`, `test_settings_service.py`, `test_settings_per_board.py` (targeted files, not the whole suite in one process — that yields known date/time pollution failures).
- **Web unit** — **125 files, 1478 pass, 7 skipped**, full `vitest run`, against `@fiestaboard/ui` built from FiestaUI#247.
- `ruff==0.16.2` check + format clean · `tsc --noEmit` clean · `eslint .` clean · `prettier --check .` clean.

Per the test-quality bar, every new test was proven non-vacuous by breaking the specific production behavior it covers. Six mutations, observed output:

**1. Remove the `code62_glyph` normalization and the Note guard from `src/devices.py`:**

```
FAILED test_invalid_glyph_falls_back_to_degree
  AssertionError: assert 'sparkle' == 'degree'
FAILED test_note_hardware_always_draws_a_heart
  AssertionError: note
  assert 'degree' == 'heart'
  → 2 failed, 4 passed
```

**2. Drop `code62_glyph=data.get(...)` from `BoardInstance.from_dict`** (the setting silently vanishes on every save — the failure mode a whitelist-style `from_dict` invites):

```
FAILED test_survives_a_save_and_reload
  AssertionError: assert 'degree' == 'heart'
  → 1 failed, 5 passed
```

**3. Drop the Note guard from `_resolve_code62_glyph` in the HTML renderer:**

```
FAILED test_note_substitutes_degree_with_heart
FAILED test_note_ignores_a_stale_flagship_preference
  assert '>°</span>' not in '<!DOCTYPE h...body></html>'
  → 2 failed, 30 passed
```

**4. Disable the substitution in the renderer's `_grid`:**

```
FAILED test_note_substitutes_degree_with_heart
FAILED test_flagship_draws_a_heart_when_its_flap_carries_one
FAILED test_note_ignores_a_stale_flagship_preference
  → 3 failed, 29 passed
```

**5. Drop the Note guard from `resolveCode62Glyph` in `web/src/hooks/use-board.ts`** (a stale Flagship preference reaching a Note in the picker):

```
× labels it a heart on a Note device
  Unable to find an accessible element with the role "option" and name "Heart character"
× inserts the degree symbol whichever glyph the board draws
× still shows all color buttons alongside it
  → 3 failed, 3 passed
```

**6. Make the package's `messageToGrid` ignore `code62Glyph`** (rebuilt and re-linked so the app really ran against it):

```
× draws a heart on a Flagship whose flap carries one
  AssertionError: expected '°' to be '♥'
  → 1 failed, 8 passed
```

Note that mutation 6 leaves *"draws a degree on a Flagship that was not told which flap it has"* passing, which is the point: that test guards the no-change-for-existing-installs promise and must not move when the new behavior breaks.

## Not done here

- **Not every preview surface** threads the glyph. Wired: the dashboard board, the page-builder editor preview and its toolbar, and the wizard's live preview — the surfaces where a user looks at their own board's content. Still on the package default (`degree`) and worth a follow-up: `inline-board-preview.tsx` (which also hardcodes `boardType="black"`), the `page-grid-selector` thumbnails, and the plugin previews on the Integrations routes.
- **MCP page previews** have no single board to read: a Page belongs to a device *shape*, not to one board. `_code62_glyph_for_device` uses the setting only when every configured board of that shape agrees, and falls back to `degree` when they disagree, rather than guessing. A per-board MCP preview would need a board argument on that resource.
- **No Playwright coverage** of the new Settings → Hardware control.
- The wrappers in `web/src/components/board-display.tsx`, `scaled-board-display.tsx` and `static-board-display.tsx` needed no change: they are `Omit<UI…Props, …>` + prop spread, so `code62Glyph` flows through as soon as the package version carries it.


jeffredodd:
## Is the setting actually honored, or does it just exist in the UI?

Honored. Verified end to end against a container built from this branch
(`fiestaboard:wt62`, prod target, isolated stack on `:4499` with its own mock
board + mock cloud), not by reading the diff. Evidence per point:

| # | Claim | Verdict | Evidence |
|---|---|---|---|
| 1 | Persists and survives reload | ✅ | `PUT /api/settings/board` with `code62_glyph: "heart"` → `GET` returns `heart`; `POST /api/settings/board/add` with `code62_glyph: "heart"` also persists. `from_dict` has no allow-list — it reads the key explicitly and `to_dict` is `asdict`, so nothing is dropped. |
| 2 | Web preview honors it | ✅ | On a Flagship with a `72°F` page, tile `char-tile-0-2` renders `♥` with the board set to heart and `°` with it set to degree — in **both** the page editor preview and the dashboard's active-page board. |
| 3 | Accessible name agrees with the tile | ✅ | Dashboard board `role="img"` name is `"Board display: 72♥F"` on a heart board and `"Board display: 72°F"` on a degree board. Structurally guaranteed: the package's `messageToText` and `messageToGrid` both call `resolveCode62Glyph` and share one substitution. |
| 4 | The wire never changes | ✅ | `BoardChars.text_to_codes` maps all of `°`, `❤`, `♥` to `62`; `"72°F"`, `"72❤F"`, `"72♥F"` all encode to `[33, 28, 62, 6]`. End to end through the mock board, the same page sent with the board at heart and at degree delivered byte-identical `[33, 28, 62, 6, 0, 0]`. |
| 5 | Note / Note Array unaffected | ✅ | `resolveCode62Glyph` (app), `_resolve_code62_glyph` (Python) and the package copy all force heart for `note`/`note_array`; a Note board with a stale `code62_glyph: "degree"` still draws `♥`. The Settings card renders **no** code-62 control for a Note board (`Degree flap`/`Heart flap` buttons absent from its button list). |
| 6 | Python preview path honors it | ✅ | In-container: `render_board_html("72°F", device_type="flagship", code62_glyph="heart")` emits `♥`, `"degree"` emits `°`, omitted emits `°`, and `device_type="note"` emits `♥` even with `"degree"` passed. `_code62_glyph_for_device("flagship")` read `heart` off the live settings service. |

### Two things worth a look (neither blocks the fix)

1. **`web/src/hooks/use-effective-code62-glyph.ts` is dead code.** Nothing
   imports `useEffectiveCode62Glyph` — not app code, not tests. Its sibling
   `useEffectiveBoardColor` *is* used, by `integrations._index.tsx` and
   `integrations.$pluginId.tsx`. ESLint does not flag it, so CI stays green on
   it. Related asymmetry at `active-page-display.tsx:696`: `boardType` falls
   back to `getEffectiveBoardColor(boardSettings)` when `currentBoard` is
   undefined, but `code62Glyph` has no such fallback and resolves to `degree` —
   which is what the hook would have fixed. Only observable during the brief
   window before board settings load, so cosmetic; but it is either "wire the
   hook up" or "delete it".
2. **Plugin previews cannot honor the setting yet.** `PluginCard` and
   `BoardShowcase` in `@fiestaboard/ui` take `boardType` but no `code62Glyph`,
   so the Integrations marketplace boards always draw `°`. Latent only — no
   shipped plugin manifest has `°` in a `teaser` or `previews`. Fixing it needs
   a FiestaUI change.

## The E2E failure

`multi-board.spec.ts:416` did `getByText("Flagship")`. The new wizard help copy
("Character code 62 shows a degree sign on Flagships made before 2026…") also
contains "Flagship", so the locator resolved to two elements and failed strict
mode:

```
strict mode violation: getByText('Flagship') resolved to 2 elements:
    1) <span …>Flagship</span> aka getByRole('button', { name: 'Flagship 6 × 22 characters' })
    2) <p …>Character code 62 shows a degree sign on Flagship…</p>
```

The copy is correct; the locator was too loose. Both board-type pills are now
addressed by role + accessible name, which folds the two dimension assertions
into the same locators. Swept the whole suite for the same shape — the only
other hit is `getByText("Flagship Only")` in `regression/pages-list.spec.ts`,
which names a page and cannot collide.

## New coverage

`web/tests/code62-glyph.spec.ts` — one behavior per test:

**Board-scoped (9):**

1. editor preview draws `°` when the flap is set to degree
2. editor preview draws `♥` when the flap is set to heart
3. a board stored without the setting draws the `°` it always drew
4. the accessible name announces the same glyph the tiles draw
5. toggling the flap in Settings changes what the preview draws
6. the flap chosen in Settings survives a reload
7. a Note board is offered no code-62 choice
8. a Note board draws `♥` even with a degree preference stored
9. changing the flap does not change what is sent to the board

**Setup wizard (2)** — the wizard swatch had a `data-testid` that nothing
referenced, so the first surface most owners will touch had no coverage at all:

10. the flap chosen in the wizard is saved with the board
11. the wizard offers no code-62 choice once Note is selected

### Non-vacuity evidence

Per the test-quality bar, every test was proven non-vacuous by breaking the
production behavior it covers, rebuilding the image, and observing the failure.
Observed output:

```
M1  use-board.ts resolveCode62Glyph ignores the stored value
    ✘ editor preview draws a heart …          Expected: "♥"  Received: "°"
    ✘ accessible name announces the same …    Expected: "♥"  Received: "°"
    ✘ toggling the flap in Settings …         Expected: "♥"  Received: "°"
    (6 passed — incl. the wire test, correctly: M1 is display-only)

M2  resolveCode62Glyph always returns "heart"
    ✘ editor preview draws a degree sign …    Expected: "°"  Received: "♥"

M3  note guard removed from BOTH mirrors (use-board.ts AND the copy inside the
    built @fiestaboard/ui bundle — breaking either one alone still draws ♥,
    which is itself worth knowing: the rule is defended twice)
    ✘ a Note board draws the heart …          Expected: "♥"  Received: "°"

M4  BoardInstance.from_dict defaults code62_glyph to "heart"
    ✘ a board stored without the setting …    Expected: "degree"  Received: "heart"

M5  from_dict stops reading the stored key
    ✘ the flap chosen in Settings survives …  aria-pressed Expected: "true"  Received: "false"

M6  the Settings control renders for every device type
    ✘ a Note board is offered no choice …     toHaveCount Expected: 0  Received: 1

M7  BoardChars.DEGREE = 63
    ✘ changing the flap does not change …     Expected: 62  Received: 63

M8  the preference perturbs the payload while leaving code 62 alone
    ✘ changing the flap does not change …     deep equality failed
       (both `[GLYPH_COL] === 62` assertions still passed — so this proves the
        "identical bytes across settings" clause specifically, not just "62")
```
```
M9  the wizard stops sending code62_glyph with the board it saves
    ✘ the flap chosen in the wizard is saved …  Expected: "heart"  Received: "degree"

M10 the wizard offers the choice for every device type, Note included
    ✘ the wizard offers no code-62 choice …     toHaveCount Expected: 0  Received: 1
```

## Full suite

```
570 passed, 23 skipped, 0 failed  (11.9m, --workers=1 --retries=0)
```

Run against a prod-target image built from this branch, on an isolated stack
(`:4499`, own mock-board and mock-cloud containers, own docker network) so it
could not borrow state from the shared dev container. **No reruns were needed
and no retries were enabled** — the three known flakes in this suite
(multi-board delete, pages-edit, schedule-form select) all passed first try, so
there was nothing to distinguish as flaky vs. real. `tsc --noEmit` reports 0
errors; `prettier --check` and `eslint` are clean on both changed files.

PR #1669, opened 2026-08-22
fix(integrations): draw plugin previews with the board's own code-62 flap

### LINKED ISSUE #1666: Plugin previews ignore the per-board code-62 flap setting
Follow-up to #1657. Not blocking that fix — filed so the gap is tracked rather than lost.

## What

`BoardInstance.code62_glyph` lets an owner say whether their Flagship's character-code-62 flap carries a degree sign or a heart, and the dashboard, editor preview, and wizard all honor it. **Plugin previews do not.**

`PluginCard` and `BoardShowcase` in `@fiestaboard/ui` accept `boardType` but no `code62Glyph`:

- `src/components/plugin/plugin-card.tsx:42` — `boardType?: "black" | "white"`, passed to `ScaledBoardTeaser` at `:106`
- `src/components/plugin/board-showcase.tsx:49` — same shape, passed through at `:112`

With no prop to thread, every board rendered on the Integrations page falls back to the package default and draws `°`, regardless of the board's setting.

## Why it isn't urgent

Latent today: no shipped manifest puts code 62 in a `teaser` or `previews`. Verified across all in-repo `plugins/*/manifest.json` plus `plugin-registry.json` and `plugin-previews.json` — zero occurrences of `°`, `❤`, or `♥`.

The first plugin author who puts a temperature in a teaser makes it visible, on exactly the boards the original bug was about.

## Fix

1. **FiestaUI:** add `code62Glyph` to `PluginCardProps` and `BoardShowcaseProps`, thread it to `ScaledBoardTeaser` / the inner board. Release.
2. **Here:** pass the resolved glyph from the Integrations routes, the same way `boardType` is already passed. `getEffectiveCode62Glyph(boardSettings)` in `web/src/hooks/use-board.ts` is the helper — `page-builder.tsx:257` shows the composition.
3. Add a manifest fixture that puts code 62 in a teaser, so the path stops being untestable.

jeffredodd:
Fixes #1666.

## What

`BoardInstance.code62_glyph` lets an owner say whether their Flagship's character-code-62 flap carries a degree sign or a heart (#1657), and the dashboard, editor preview and wizard all honor it. The Integrations page did not — `PluginCard` and `BoardShowcase` were handed `boardType` and nothing else, so every marketplace teaser and every hero preview drew `°` regardless of the board's setting, on exactly the boards #1657 was about.

- **Package half** (Fiestaboard/FiestaUI#264, released as `@fiestaboard/ui` 5.4.1): `code62Glyph` threaded through `PluginCard` → `ScaledBoardTeaser` → `BoardTeaser` and through `BoardShowcase` → `StaticBoardDisplay`
- **Here**: bump to 5.4.1 and pass the resolved glyph from both Integrations routes, the same way `boardType` already is

New `useEffectiveCode62Glyph()` sits beside `useEffectiveBoardColor()` (same import-cycle reason for its own module): it reads the board being managed and falls back to board settings while that is still loading, resolving the device alongside the preference — a Note is heart hardware whatever a stale Flagship preference says.

A `note` / `note_array` preview in the showcase still draws its own heart: `StaticBoardDisplay` resolves that from the preview's `device_type`, not from this prop.

## Tests

`web/src/__tests__/integrations-plugin-preview-code62.test.tsx` mounts both routes against MSW fixtures. The registry entry it mocks is the **manifest fixture the issue asks for** — no shipped manifest puts code 62 in a `teaser` or `previews`, which is the only reason this was latent, so the test supplies one (`teaser: "52 °F CLEAR"`).

Written first and watched fail — the two heart cases failed, the two degree guards passed:

```
× draws a heart in the marketplace teaser when the board's flap carries one
  → Unable to find role="img" and name `/52 ♥F CLEAR/`
✓ keeps the degree in the marketplace teaser for a board saved before the setting
× draws a heart in the plugin detail showcase when the board's flap carries one
  → Expected element to have text content: 52♥FCLEAR / Received: 52°FCLEAR
✓ keeps the degree in the plugin detail showcase for a board saved before the setting
```

The marketplace assertion reads the strip's `role="img"` name rather than its tiles: the tiles are `aria-hidden`, so the name is all a screen reader gets, and it comes from the same substitution — a strip drawing ♥ while announcing "degree" would be a text alternative for a different image (WCAG 1.1.1).

After the fix: 4/4 pass. Full web suite on this branch with 5.4.1 installed: **128 files, 1498 passed, 7 skipped**. `typecheck`, `eslint` and `prettier --check` clean.

FiestaUI side: 8 new mounted tests (5 of which failed before that fix), `release:test` 242/242, vitest 131/131.

## Note on #1668

The release bot opened #1668 for the same 5.4.1 bump. This PR carries the identical bump because the app change does not compile without it, so #1668 is redundant with this one — closing it in favour of this.

### DIFF
--- a/web/app/routes/integrations.$pluginId.tsx
+++ b/web/app/routes/integrations.$pluginId.tsx
@@ -36,6 +36,7 @@ import { toast } from "sonner";
 
 import Link from "@/components/smart-link";
 import { useEffectiveBoardColor } from "@/hooks/use-effective-board-color";
+import { useEffectiveCode62Glyph } from "@/hooks/use-effective-code62-glyph";
 import { useParams, useRouter } from "@/hooks/use-router";
 import { useTranslations } from "@/i18n/translations";
 import { api } from "@/lib/api";
@@ -49,6 +50,7 @@ export default function PluginDetailPage() {
   const router = useRouter();
   const queryClient = useQueryClient();
   const boardColor = useEffectiveBoardColor();
+  const code62Glyph = useEffectiveCode62Glyph();
   const pluginId = params.pluginId as string;
   const [addInstanceOpen, setAddInstanceOpen] = useState(false);
   const [instanceLabel, setInstanceLabel] = useState("");
@@ -158,6 +160,7 @@ export default function PluginDetailPage() {
             previews={previews}
             previewLabel={t("boardPreviewLabel", { name: entry?.name ?? pluginId })}
             defaultBoardType={boardColor}
+            code62Glyph={code62Glyph}
             labels={{
               flagship: t("deviceFlagship"),
               note: t("deviceNote"),
--- a/web/app/routes/integrations._index.tsx
+++ b/web/app/routes/integrations._index.tsx
@@ -277,6 +277,7 @@ import { asJSONSchema, SchemaForm } from "@/components/plugin-settings";
 import Link from "@/components/smart-link";
 import { useDepsChanged } from "@/hooks/use-deps-changed";
 import { useEffectiveBoardColor } from "@/hooks/use-effective-board-color";
+import { useEffectiveCode62Glyph } from "@/hooks/use-effective-code62-glyph";
 import { useSearchParams } from "@/hooks/use-router";
 import { useTranslations } from "@/i18n/translations";
 import type { PluginInfo, RegistryEntry } from "@/lib/api";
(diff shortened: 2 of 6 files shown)


jeffredodd:
Superseded by #1670 — same commit, rebased onto main after #1667 landed. This PR's branch conflicted with main (`web/package.json` / `web/package-lock.json`), which is why GitHub never dispatched the `pull_request` CI run: with no mergeable merge ref, only the `pull_request_target` workflows fired.

Issue #1671, opened 2026-08-22
Plugin requirements.txt is never installed — plugins with third-party deps are broken on install


jeffredodd:
## Summary

A plugin's `requirements.txt` is never installed. Nothing in the install path, the loader, or the container build reads it — the only consumer in the tree is `scripts/scan_registry.py`, which feeds it to `pip-audit` for vulnerability scanning:

```
$ grep -rn 'requirements.txt' src/ scripts/ Dockerfile
scripts/scan_registry.py:234:    req_file = os.path.join(plugin_dir, "requirements.txt")
Dockerfile:21:COPY requirements.txt .          # the platform's own, not a plugin's
```

So any plugin that needs a third-party package is broken the moment it is installed, and it fails at `fetch_data()` time with a message the user cannot act on.

## Evidence

Loading all 58 discoverable plugins through the real `PluginLoader` and calling `fetch_data()` with schema defaults:

| plugin | result |
|---|---|
| `volcano` | `UNAVAIL` — `No module named 'feedparser'` |
| `network_speed` | `UNAVAIL` — `speedtest-cli package not installed` |

Cross-checking declared deps against the runtime — **2 of 2** plugins that declare any dependency are affected:

```
plugin                   declared deps    missing at runtime
network_speed            speedtest-cli    speedtest-cli (import speedtest)
volcano                  feedparser       feedparser (import feedparser)

2 plugins declare deps; 2 have deps MISSING from the runtime
```

Blast radius is small today only because almost every plugin sticks to `requests` + stdlib. The mechanism is entirely absent, so it will bite harder as the registry grows.

## Why it is invisible today

Per-plugin CI installs `fiestaboard-core/requirements.txt` **and** the plugin's own dev requirements, so the dep is present in the test environment. It goes missing only on a real install. This is the same class of gap as Fiestaboard/fiestaboard-plugin--star-trek-quotes#10, where CI fabricated a missing data file and the suite passed against a layout no user has.

## Notes on a fix

Worth deciding deliberately rather than defaulting — installing arbitrary packages from a registry plugin is a supply-chain surface, and the plugins run in-process:

1. **Install at plugin-install time** — `pip install -r requirements.txt` in `src/plugins/sources.py`. Simplest, but executes third-party package installs (including build hooks) on the user's box, and there is no lockfile or hash pinning today.
2. **Declare deps in `manifest.json` against an allowlist** the platform pre-installs, so the set of installable packages is reviewed rather than arbitrary.
[...]

PR #1676, opened 2026-08-22
feat(plugins): validate plugins at install time instead of after the fact

jeffredodd:
Follow-up to #1675. That PR added a daily sweep, which finds a broken plugin a day late and infers severity from a regex over source. This moves the check to the boundary: a plugin that cannot work is never installed in the first place.

## The declaration

`manifest.json` gains an optional `data_files`:

```json
{
  "id": "star_trek_quotes",
  "data_files": ["quotes.json"]
}
```

Paths are relative to the plugin directory. Absolute paths and `..` segments are rejected — a plugin reaching outside its own directory only resolves in the bundled `plugins/<id>/` layout and breaks once installed to `data/external_plugins/<id>/`. That is exactly the Star Trek Quotes regression ([plugin#10](https://github.com/Fiestaboard/fiestaboard-plugin--star-trek-quotes/pull/10)).

## The checks

New `src/plugins/install_check.py` answers one question — *can this plugin possibly work here?* It deliberately does not ask whether it is **configured**; that is the user's business and comes later.

- declared data files are present
- the declaration is well-formed and stays inside the plugin directory
- `requirements.txt` entries are importable (#1671)
- `__init__.py` / `manifest.json` exist

**Severity tracks whether the plugin can work, not whether the author remembered to declare anything.** This is the subtle part, and my first cut got it wrong: I initially made all undeclared-file findings warnings, which downgraded the Star Trek case — *which declared nothing* — from broken to advisory. The rule that survives:

| situation | severity |
|---|---|
| path escapes the plugin directory | **error** (can never resolve once installed) |
| file is read but missing | **error** (broken now, declared or not) |
| file is read, present, undeclared | advisory (works, but unprotected) |

## Enforcement

| when | what happens |
|---|---|
| **install** | a fresh install that fails is **refused** and the clone removed |
| **update** | reported via `GET /plugins/errors`, but **left installed** — an update must not pull a working board out from under someone |
| **startup** | re-checked; findings surface in `GET /plugins/errors` |
| **registry** | `validate_plugins.py --strict` escalates the advisory |
| **daily** | the sweep from #1675 now delegates here instead of carrying its own copy (net −233 lines in that script) |

## Verification

Against the **real instance**, the installed `star_trek_quotes` — which declares nothing — is now reported through `/plugins/errors`:

```
star_trek_quotes:
  - __init__.py reads 'src/utils/star_trek_quotes.json' from outside the plugin
    directory (2 levels up). That path only resolves in the bundled
    plugins/<id>/ layout and breaks once the plugin is installed.
  - __init__.py reads 'quotes.json' but no such file ships with the plugin.
    Every variable it provides will render '???'.
```

Swept across all 51 published plugin repos: **2 blocking** (`volcano`, `network_speed` — both #1671), **2 advisory** (present-but-undeclared reads, incl. one in `national_day`), **zero false positives**.

## Tests

87 cases across three files, every check with a matching negative case.

**Mutation-tested** — each of these turns tests red:

| mutation | tests failed |
|---|---|
| don't clean up a rejected install | 1 |
| block updates too (violates the policy) | 2 |
| skip the data-file check | 5 |
| skip the dependency check | 2 |
| allow `..` traversal in declarations | 3 |

**Three existing tests changed, deliberately.** `test_plugin_sources.py` mocked `clone_or_update_repo` to return `(True, "")` while writing nothing — encoding "a successful clone is sufficient to install." That contract is now false on purpose. Their fake clone leaves a plausible plugin behind, which is what a real clone does, and a **new** test covers the rejection path they used to occupy. This is a corrected contract, not a weakened assertion.

**A/B against `origin/main`** on the same test selection, same container: 1155 → 1194 passing, error set byte-identical, no regressions. (The 78 errors in both are pre-existing artifacts of the isolated test dir, not of either branch.)

`ruff check` / `ruff format --check` clean; full `npm run docs:lint` clean.

## Known limitation

The declaration is opt-in, so it does nothing for the existing fleet until plugins adopt it — I verified this rather than assuming: the installed `star_trek_quotes` produced **zero** findings from the declaration path alone. The source scan is what covers those plugins, and it is a regex, not an AST pass. That is why it can only ever *warn* about a file that is present, and why the real-corpus run above matters more than the unit tests for judging its precision.

Phase 2 (a typed `unconfigured` vs `error` reason on `PluginResult`) removes the remaining guesswork and fixes the `???`-means-three-things ambiguity in the UI. Separate PR.

### DIFF
--- a/docs/internal/development/PLUGIN_DEVELOPMENT.md
+++ b/docs/internal/development/PLUGIN_DEVELOPMENT.md
@@ -746,7 +746,20 @@ This is not hypothetical: the Star Trek Quotes plugin shipped this way for
 months, serving `???` to every user, because its data file was never committed
 and its fallback pointed at the platform copy.
 
-Two rules follow:
+Three rules follow:
+
+- **Declare your data files** in `manifest.json`, so a broken install is
+  rejected before the plugin ever runs:
+
+  ```json
+  {
+    "id": "star_trek_quotes",
+    "data_files": ["quotes.json"]
+  }
+  ```
+
+  Paths are relative to your plugin directory. Absolute paths and `..`
+  segments are rejected — a plugin may only declare files it ships.
 
 - **Commit your data files.** Do not generate or symlink them in CI. If your
   test workflow creates a file the shipped plugin lacks, your suite is testing
@@ -757,8 +770,17 @@ Two rules follow:
   beyond the standard library plus the platform's own dependencies will be
   missing at runtime. Vendor it or do without.
 
-`scripts/plugin_health_sweep.py` enforces all of this against every registry
-plugin daily. Run it locally before publishing:
+### How this is enforced
+
+| when | what happens |
+|------|--------------|
+| **Install** | A plugin that declares a file it does not ship, reads a file that is missing, or needs a package FiestaBoard lacks is **refused**, and the clone is removed. |
+| **Update** | The same failure is reported through `GET /plugins/errors` and the Integrations page, but the plugin is **left installed** — an update should not silently pull a working board out from under someone. |
+| **Startup** | Every installed plugin is re-checked; findings appear in `GET /plugins/errors`. |
+| **Registry review** | `validate_plugins.py --strict` escalates the advisory "reads a file it does not declare" into a failure. |
+| **Daily** | `scripts/plugin_health_sweep.py` sweeps every registry plugin and files an issue. |
+
+Run the checks locally before publishing:
 
 ```bash
 docker compose -f docker-compose.dev.yml exec fiestaboard \
--- a/scripts/plugin_health_sweep.py
+++ b/scripts/plugin_health_sweep.py
@@ -45,10 +45,8 @@ Exit codes:
 """
 
 import argparse
-import importlib.util
 import json
 import os
-import re
 import sys
 import traceback
 from pathlib import Path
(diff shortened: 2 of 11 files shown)


jeffredodd:
Reviewed. The design is right — validating against the tree that actually landed, refusing a fresh install but never yanking an installed plugin out from under a running board, and keeping the regex scan advisory except where the plugin is provably broken. The path-injection barrier around the rejection `rmtree` is appropriately paranoid.

I pushed one fix (3c068f7) rather than merge a latent trap.

**`_is_importable` guessed the import name from the distribution name**, via an 8-entry override table plus dash-to-underscore. That guess feeds `check_dependencies`, whose errors are *blocking* — and a blocking error on a fresh install refuses the install and deletes the clone. FiestaBoard ships two packages the guess gets wrong, neither in the table:

| distribution | imports as | guess | `find_spec` | shipped? |
|---|---|---|---|---|
| `finnhub-python` | `finnhub` | `finnhub_python` | ❌ | ✅ yes |
| `paho-mqtt` | `paho.mqtt` | `paho_mqtt` | ❌ | ✅ yes |

Verified in the running container. A plugin declaring either would have been rejected for a dependency that is in fact present. Installed metadata is keyed by the name pip installs under and normalises per PEP 503, so it needs no guess — it now asks that first and keeps `find_spec` as the fallback.

The two true positives are unaffected: `feedparser` and `speedtest-cli` are absent under both methods, and `test_distribution_name_is_mapped_to_import_name` still passes. Added a parametrised regression test; confirmed it fails on the pre-fix code for the right reason (`assert ["requires Python package 'finnhub-python', which is not available..."] == []`) and passes after. Full affected suite: **226 passed**, ruff clean.

Two things I am *not* blocking on, worth a follow-up:

1. **The source scan cannot tell a read from a write.** A plugin that writes a cache file next to `__init__.py` (`Path(__file__).parent / "cache.json"`) trips the "reads `'cache.json'` but no such file ships" branch on a fresh install, which refuses it. Zero occurrences across the 51 published repos today, so this is latent rather than live — but the failure is unrecoverable from the UI, and the regex has no way to distinguish the two.
2. **Dead code in `_declared_requirements`:** `if "dev" in req.name: continue` can never fire, because `rglob("requirements.txt")` only ever matches that exact filename. `test_dev_requirements_are_ignored` therefore passes for the wrong reason — `requirements-dev.txt` is never globbed in the first place. Behaviour is correct; the guard and the test are both misleading.

PR #1776, opened 2026-08-28
feat: FiestaPi HDMI output — one-click enable from the app, kiosk at /p/display

jeffredodd:
Stacked on #1775 (FiestaPanel). A FiestaPi can drive a TV/monitor over its HDMI port, showing the reserved **`/p/display`** FiestaPanel URL — and **existing deployed Pis enable it with one click in the app**, no SSH, no re-flash.

## How users turn it on

**Settings → Hardware → FiestaPanel → "HDMI output on this FiestaPi"** (a switch, shown only on pi-profile installs with the updater sidecar). Flipping it installs a minimal kiosk (cage + Chromium) on the host and boots it onto the connected screen. The screen then follows whichever panel has **Display output** turned on (single-holder role from #1775) — re-pointed entirely from the app.

## How it reaches the fleet

`fiestaupdater` is the one component on every deployed Pi that is host-root-capable (Docker socket) **and self-refreshes** (the boot service `docker compose pull`s both images every boot). This PR gives it three fixed verbs:

- `GET /hdmi/status` (no auth, like `/last-update`) — persisted state for the UI
- `POST /hdmi/enable` / `POST /hdmi/disable` (bearer-token) — queue the host change

The worker runs the baked-in retrofit script *inside the host's own namespaces*: a throwaway privileged helper container (`--pid=host`, using the service's already-present image, which ships `nsenter`) enters PID 1's namespaces and executes the script from stdin. **The API surface never accepts scripts or arguments from callers — only these two fixed actions exist**, and the Docker socket the sidecar already holds is host-root-equivalent anyway, so this adds no new trust class. If a fleet sidecar predates the verbs, the app maps the 404 to "reboot the Pi to update it, then try again."

The app proxies the verbs behind its own auth at `GET/POST /settings/hdmi-kiosk`, with install-progress polling and full i18n (14 locales).

## Three install paths, one source of truth

1. **In-app switch** (primary, works on the whole existing fleet)
2. **Fresh images**: pi-gen sub-stage `02-hdmi-kiosk` bakes the kiosk in, gated by `ConditionPathExists=/boot/firmware/fiestapi-hdmi.txt` (flash-time opt-in; the switch works there too)
3. **SSH fallback**: `curl … scripts/fiestapi-hdmi-setup.sh | sudo bash` (idempotent, `--disable` supported)

All three share one script: `tests/test_fiestapi_hdmi_sync.py` pins the sidecar's baked copy, the image stage's unit/wait files, and the canonical `scripts/` copy **byte-for-byte** so they can never drift.

Kiosk service details: unprivileged `kiosk` user on tty7 per cage's systemd guide, `WantedBy=multi-user.target` (Pi OS Lite never reaches graphical.target), health-gated start so first boots never paint a connection error, hard opt-in via the flag file so a disabled kiosk costs zero RAM on a 1 GB Pi 3B.

## Testing

- **Sidecar**: bats suite grown 27 → 33 (status persistence, auth, helper invocation shape `--privileged --pid=host nsenter`, `--disable` pass-through, missing-script 500) — all green in a bats container.
- **App**: 9 new endpoint tests (profile/sidecar gating, token forwarding, verb routing, stale-sidecar 409, validation) + settings-card tests for the switch (hidden when unsupported, POSTs `{"enabled": true}`) — green, tsc/eslint/prettier/ruff/docs-lint clean.
- **Sync**: 3 byte-identity tests across the three install paths.
- ⚠️ **Hardware pass still required**: flash a CI-built image AND exercise the in-app switch on an already-flashed Pi (that's the flow this PR exists for). Acceptance: switch → progress → kiosk appears on TV → instruction card → designate panel → screen re-points; disable → Chromium gone after reboot; sidecar-too-old path shows the reboot hint.


jeffredodd:
Added the deployed-fleet story: `fiestaupdater` only updates the app container, never the host OS, so existing Pis can't get the kiosk from an app update. This PR now also ships `scripts/fiestapi-hdmi-setup.sh` — a one-command SSH retrofit (`curl … | sudo bash`, idempotent, `--disable` to turn off) that installs the identical pieces the image bakes in, plus a sync test (`tests/test_fiestapi_hdmi_sync.py`) pinning the script's embedded systemd unit and wait script byte-for-byte to the image stage files so fresh flashes and retrofits can never drift apart. Both paths share the same hard opt-in: the kiosk only runs while `fiestapi-hdmi.txt` exists on the boot partition.

Issue #1746, opened 2026-08-24
GET /silence-status runs a data migration under the wrong lock


jeffredodd:
### Why
The GET handler calls `migrate_silence_schedule_to_utc()`
(`src/api_server.py:4266`) — a config write triggered by a read. The migration also
takes the class-level singleton-construction lock instead of the file lock
(`src/config_manager.py:1338`), blocking `ConfigManager` construction in other
threads while doing disk I/O.
### What
Run the migration once at startup; take `_file_lock`; make the GET pure.
### Acceptance
GET performs no writes; migration idempotence test still green.

_Part of #1730 (Phase 1)._

PR #1801, opened 2026-09-03
fix(silence): per-board silence schedules

### LINKED ISSUE #1788: Silence Settings are global vs Board specific
FiestaPi has the ability to manage multiple, separate boards. However, the Silence settings are global, which means the following:
1. Each separate Vestaboard is forced to have the same silence time period (undesirable)
2. The settings for the 'Page' to be displayed during the silence period can not be different for each board being managed. This is a problem if you have 1 note, and 1 flagship - the Page selection can only be of one size.

**Expected behavior**
Silence settings are Board specific, not global.

**Environment**
- FiestaBoard version: 8.33.9
- Hardware: Raspberry Pi 3

jeffredodd:
Closes #1788.

Silence settings were install-wide, so a mixed Note + Flagship setup could not be configured correctly.

- Resolve the silence schedule per board in the config layer; an unconfigured board inherits the install-wide schedule
- Board-scoped silence API with page/board size validation
- Per-board silence delivery in the display engine
- Scope the silence settings UI to the selected board (`useCurrentBoard()`)
- Filter the silence page picker with `pagesCompatibleWithBoard` — it previously offered pages of any size
- Docs: per-board resolution rule and `GET /silence-status?board_id=`

Also flags the dead `SILENCE_SCHEDULE_START_TIME` / `_END_TIME` env vars — they write `general.silence_schedule_*`, which nothing in `src/` or `web/` reads, so the documented examples were a no-op.

Rebased onto current `main`. Two of the merges it picked up matter here:

- **#1817** rewrote `check_and_send_for_board` (silence evaluated above the pause check, `rt.last_silence_mode_active` latched on every exit path). Kept intact, and it is now what the per-board boundary detector compares against.
- **#1746** moved the silence migrations off `GET /silence-status` into `_run_startup_migrations`, and made `ConfigManager._file_lock` an `RLock`. The per-board seeding migration is added to the startup runner (before the UTC pass, so it converts the seeded windows in the same boot) and `GET /silence-status` stays a pure read. The `RLock` change also affects BLOCKER 2 below.
- **#1813** added `DisplayService.board_init_errors`; the per-board silence snapshot sits alongside it.

---

## Review fixes

Four defects found in code review. Every piece of the original PR was correct in isolation; the composition was not, and 24 green tests said nothing about it.

### BLOCKER 1 — single-board installs could not change silence settings at all

The seeding migration gave **every** configured board an explicit copy of the install-wide window, single-board installs included. The engine always resolves per board (`check_and_send_for_board` is called with the primary board's id, never `None`). But the UI wrote the **install-wide** layer whenever `boards.length === 1`. The seeded copy then shadowed every later save key by key:

```
by_board after migration: {"board-1": {"enabled": true, "start_time": "20:00+00:00", ...}}
--- user turns silence OFF and moves the window to 22:30 in the UI ---
install-wide layer after save : enabled=False start=22:30+00:00
what the ENGINE resolves      : enabled=True  start=20:00+00:00 end=07:00+00:00
what the UI form shows        : enabled=False start=22:30+00:00
```

Also reachable without the migration: delete a board from a 2-board install and `isMultiBoard` flips false while the survivor keeps its `by_board` entry.

**Design choice: (b) — the writer always targets the board layer.**

The reader is board-scoped on every install and cannot be anything else: `check_and_send_active_page()` resolves `get_primary_board_id()` and hands it to `check_and_send_for_board`. So the only way to make the write layer and the read layer agree *in every case* is to make the writer board-scoped too, unconditionally.

Why not the others:

- **(a) don't seed `by_board` for single-board installs** does not fix the 2 → 1 deletion path: pruning the *removed* board still leaves the *survivor's* entry, `isMultiBoard` still flips false, and the shadowing returns. It also depends on an invariant ("a single-board install never has a `by_board` entry") that any board-scoped write from MCP or curl silently breaks.
- **(c) make resolution prefer the install-wide layer when there is exactly one board** puts board-count logic on the *read* side, so adding a second board would silently change board 1's effective schedule. It also forces `resolve_silence_schedule` — today a pure function of the feature dict — to read board settings, which is exactly the re-entrancy that produced BLOCKER 2.

Implemented on both sides of the wire, because a client-side convention alone cannot guarantee "in every case":

- `web/src/components/settings/silence-schedule.tsx` drops the `isMultiBoard` conditional and reads **and** writes `by_board[currentBoardId]` unconditionally.
- `PUT /settings/silence-schedule` enforces the same invariant for clients that are not our UI: an install-wide write on a single-board install prunes that board's override instead of letting it win. On one board there is no meaningful difference between "the install default" and "this board", so a divergent override there can only shadow the save. Multi-board installs are untouched — there the overrides are the whole point.

#### Follow-on, caught by E2E: `GET /silence-status` was still the wrong layer

Making the form write the board layer fixed the engine, but the status endpoint with no `board_id` still returned the install-wide layer — so on a single-board install the user toggled silence off and every surface polling it unscoped kept saying on:

```
  1) [chromium] › tests/settings-full.spec.ts:158:3 › Settings – Full Coverage › toggling silence schedule saves without 404 (regression: #597)
    Error: expect(received).toBe(expected) // Object.is equality
    Expected: true
    Received: false
    Call Log:
    - Timeout 5000ms exceeded while waiting on the predicate
    > 196 |       .toBe(!originalEnabled);
```

Not just that endpoint: `SilenceImminentBanner` and `ActivePageDisplay` both scope with `boards.length > 1 ? currentBoardId : undefined`, so a single-board install read the install-wide layer for the dashboard SNOOZING overlay and the countdown banner too.

`/silence-status` is a runtime status endpoint — it returns `active`, `current_time_utc`, `next_change_utc`, `seconds_until_next_change` — so "is silence on?" with no board named has to mean "on the board you drive by default", the same rule as `_silence_active(None)`, `_board_is_paused(None)` and `SettingsService.is_paused(None)`. It now resolves the primary board and echoes the board it describes. That fixes the unscoped consumers for free. The install-wide layer is still readable as raw config via `GET /settings/all` → `silence_schedule.config`, which includes `by_board`.

`test_omitting_board_id_keeps_legacy_global_reading` encoded the contract this replaces and is rewritten as `test_omitting_board_id_reports_the_primary_board`, with a companion asserting a board with no override of its own still reads the install-wide values. The per-board E2E now reads the install-wide layer from the stored config instead of from the status endpoint — which is what it actually meant to assert.

### BLOCKER 2 — latent self-deadlock that froze the whole process

`migrate_silence_schedule_to_per_board` held `self._file_lock` across a call into `get_settings_service()`, whose constructor reads `FB_TRANSITION_STRATEGY` → `Config._get_board()` → `ConfigManager.get_board()` — the same lock. Held forever, every config read in the process blocks: total freeze, no exception, no timeout. `_configured_board_ids()` is hoisted above the `with self._file_lock`.

**Caveat, found on the second rebase:** #1746 landed mid-review and made `_file_lock` an `RLock`, so the same-thread re-entry no longer hangs and the original timeout-based test passed with the fix reverted. That test was vacuous on current `main` and has been replaced. The invariant now pinned is the durable one — *the config lock must be free while the settings service is being constructed* — proved by a probe thread that tries to acquire it from inside the callback. That fails whether the lock is reentrant or not. The hoist is kept regardless: building a service that reads config under the config lock is a deadlock waiting on `RLock` staying an `RLock`, and it also holds the lock across unrelated disk I/O.

### MAJOR 3 — eight call sites still resolved install-wide

Install-wide silence disabled, bedroom Note overridden to 22:00–07:00 — at 2am a manual send from the web UI or Home Assistant went straight through. New `_silence_active(board_id)` mirrors `_board_is_paused`: an explicit board is used as given, `None` resolves to the **primary** board rather than the install-wide layer. `Config.is_silence_mode_active(None)` keeps its legacy install-wide meaning for the ~20 fixtures that call it zero-arg.

| Call site | Board threaded through | Note |
| --- | --- | --- |
| `POST /send-message` (`api_server.py:3345`) | primary board | No board parameter exists; the endpoint drives `service.vb_client` |
| `POST /send-welcome-message` (`api_server.py:3493`) | primary board | Setup wizard; builds a fresh primary client from `Config` |
| `POST /transitions/test-live` (`api_server.py:5751`) | `request["board_id"]` | Already resolved for the client |
| `POST /transitions/restore` (`api_server.py:5845`) | `body["board_id"]` | Already resolved for the client |
| `POST /pages/{id}/send` (`api_server.py:8042`) | `board_id` param | Already routes the client with it |
| `src/mqtt/commands.py:297` | board from the MQTT payload | Falls back to primary for a plain-string payload |
| `src/mqtt/state.py:101` | primary board | See below |

Two sites genuinely have no board context: `/send-message` and `/send-welcome-message`. Neither takes a board — both write the primary board's client — so both resolve the primary board, which is the board they are about to touch.

`src/mqtt/state.py` now publishes the **primary** board's silence to Home Assistant. Every other field in that payload is primary-scoped (`active_page`, `transition_style`, `current_message`), so publishing the install-wide layer made the sensor report OFF while the board it describes was snoozing. A per-board sensor set is a separate feature; documented in `docs/features/home-assistant-control.md`.

### MINOR 4 — orphan `by_board` entries

`SettingsService.remove_board` now prunes the removed board's override via the new `ConfigManager.prune_silence_schedule_for_board`.

---

## Verbatim pre-fix failures

Every test below was written first and run against the rebased branch before any fix.

**Composition, deadlock and prune** (`tests/test_silence_per_board_composition.py`), captured against the branch as reviewed (before the second rebase, when `_file_lock` was still a plain `Lock`):

```
============================= test session starts ==============================
collected 5 items

tests/test_silence_per_board_composition.py F.FFF                        [100%]

=================================== FAILURES ===================================
_ TestSingleBoardComposition.test_single_board_install_save_is_honored_by_the_engine _
tests/test_silence_per_board_composition.py:114: in test_single_board_install_save_is_honored_by_the_engine
    assert resolved["enabled"] is False
E   assert True is False
_ TestSingleBoardComposition.test_save_after_dropping_from_two_boards_to_one_is_honored _
tests/test_silence_per_board_composition.py:169: in test_save_after_dropping_from_two_boards_to_one_is_honored
    assert resolved["enabled"] is False
E   assert True is False
_ TestBoardRemovalPrunesOverrides.test_remove_board_prunes_its_silence_override _
tests/test_silence_per_board_composition.py:195: in test_remove_board_prunes_its_silence_override
    assert "board-2" not in by_board
E   AssertionError: assert 'board-2' not in {'board-1': {'enabled': True, 'start_time': '01:00+00:00', 'end_time': '02:00+00:00'}, 'board-2': {'enabled': True, 'start_time': '03:00+00:00', 'end_time': '04:00+00:00'}}
_ TestMigrationLockSafety.test_seeding_migration_does_not_reenter_the_config_file_lock _
tests/test_silence_per_board_composition.py:237: in test_seeding_migration_does_not_reenter_the_config_file_lock
    assert not worker.is_alive(), (
E   AssertionError: migrate_silence_schedule_to_per_board() deadlocked: it called get_settings_service() while holding ConfigManager._file_lock
E   assert not True
E    +  where True = is_alive()
E    +    where is_alive = <Thread(Thread-4 (_run), started daemon 281473271460192)>.is_alive
=========================== short test summary info ============================
FAILED tests/test_silence_per_board_composition.py::TestSingleBoardComposition::test_single_board_install_save_is_honored_by_the_engine
FAILED tests/test_silence_per_board_composition.py::TestSingleBoardComposition::test_save_after_dropping_from_two_boards_to_one_is_honored
FAILED tests/test_silence_per_board_composition.py::TestBoardRemovalPrunesOverrides::test_remove_board_prunes_its_silence_override
FAILED tests/test_silence_per_board_composition.py::TestMigrationLockSafety::test_seeding_migration_does_not_reenter_the_config_file_lock
========================= 4 failed, 1 passed in 10.52s =========================
```

The deadlock test runs the migration on a worker thread with a join timeout, so it fails red instead of hanging the suite. After the rebase onto #1746's `RLock` it no longer reproduced (see the caveat above) and was rewritten around the lock-holding probe; here it is failing against the fix reverted on the **current** base:

```
=================================== FAILURES ===================================
_ TestMigrationLockSafety.test_seeding_migration_does_not_hold_the_config_lock_across_the_board_lookup _
tests/test_silence_per_board_composition.py:258: in test_seeding_migration_does_not_hold_the_config_lock_across_the_board_lookup
    assert lock_was_free.get("free") is True, (
E   AssertionError: ConfigManager._file_lock was held across get_settings_service(); that call reads config through ConfigManager.get_board()
E   assert False is True
E    +  where False = <built-in method get of dict object at 0xffffb0041380>('free')
E    +    where <built-in method get of dict object at 0xffffb0041380> = {'free': False}.get
=========================== short test summary info ============================
FAILED tests/test_silence_per_board_composition.py::TestMigrationLockSafety::test_seeding_migration_does_not_hold_the_config_lock_across_the_board_lookup
============================== 1 failed in 2.60s ===============================
```

**Send guards** (`tests/test_silence_per_board_call_sites.py`) — each guard let the send through:

```
E   AssertionError: assert 'success' == 'blocked'
/app/tests/test_silence_per_board_call_sites.py:83: AssertionError: assert 'success' == 'blocked'
E   AssertionError: assert 'success' == 'blocked'
/app/tests/test_silence_per_board_call_sites.py:97: AssertionError: assert 'success' == 'blocked'
E   assert True is False
/app/tests/test_silence_per_board_call_sites.py:121: assert True is False
E   AssertionError: {"detail":"Page not found: page-1"}
    assert 404 == 409
/app/tests/test_silence_per_board_call_sites.py:142: AssertionError: {"detail":"Page not found: page-1"}
E   AssertionError: {"detail":"Active page not found"}
    assert 404 == 409
/app/tests/test_silence_per_board_call_sites.py:156: AssertionError: {"detail":"Active page not found"}
E   AssertionError: assert 'OFF' == 'ON'
/app/tests/test_silence_per_board_call_sites.py:190: AssertionError: assert 'OFF' == 'ON'
E   AssertionError: Expected 'send_characters' to not have been called. Called 1 times.
/usr/local/lib/python3.14/unittest/mock.py:947: AssertionError: Expected 'send_characters' to not have been called. Called 1 times.
=========================== short test summary info ============================
FAILED tests/test_silence_per_board_call_sites.py::TestApiSendGuards::test_send_message_respects_the_primary_boards_window
FAILED tests/test_silence_per_board_call_sites.py::TestApiSendGuards::test_send_welcome_message_respects_the_primary_boards_window
FAILED tests/test_silence_per_board_call_sites.py::TestApiSendGuards::test_page_send_respects_the_target_boards_window
FAILED tests/test_silence_per_board_call_sites.py::TestApiSendGuards::test_live_transition_test_respects_the_target_boards_window
FAILED tests/test_silence_per_board_call_sites.py::TestApiSendGuards::test_transition_restore_respects_the_target_boards_window
FAILED tests/test_silence_per_board_call_sites.py::TestMqttSilenceIsPerBoard::test_state_publisher_reports_the_primary_boards_silence
FAILED tests/test_silence_per_board_call_sites.py::TestMqttSilenceIsPerBoard::test_mqtt_send_message_respects_the_target_boards_window
```

`is_silence_mode_active` is replaced by a board-aware stub whose install-wide layer is explicitly **not** silent, so a guard that still passes `None` sees "not silent" and lets the send through — which is the production bug. The two 404s are proof the guard did not fire: the silence check is the first thing after the board client is resolved, so anything downstream of it (page lookup) means the request got past it.

**UI write and read halves** (`web/src/__tests__/silence-schedule-per-board.test.tsx`):

```
 ❯ src/__tests__/silence-schedule-per-board.test.tsx (10 tests | 2 failed) 8476ms
     × sends the board_id on a single-board install so the save reaches the engine 1166ms
     × seeds the form from the board's own entry on a single-board install 5030ms

 FAIL  src/__tests__/silence-schedule-per-board.test.tsx > SilenceSchedule - board scoping > sends the board_id on a single-board install so the save reaches the engine
AssertionError: expected undefined to be 'flag-1' // Object.is equality

- Expected:
"flag-1"

+ Received:
undefined

 FAIL  src/__tests__/silence-schedule-per-board.test.tsx > SilenceSchedule - board scoping > seeds the form from the board's own entry on a single-board install
AssertionError: expected 'SNOOZING' to be 'BEDTIME' // Object.is equality
```

The first replaces `expect(puts[0]).not.toHaveProperty("board_id")`, an assertion that encoded the wrong contract — "single-board installs send a byte-identical PUT body" was the safety guarantee that turned out to be the bug. This is a corrected contract, not a weakened assertion.

## Test-fixture updates (signature, not assertion)

- `tests/test_transitions_api.py` stubbed `is_silence_mode_active` with `staticmethod(lambda: False)`; the guard now passes a board, so the stubs take `board_id=None`.
- `tests/test_silence_schedule_polling.py` had one `side_effect = lambda: silence["active"]` for the same reason.
- `_silence_state_changed` now falls back to the runtime's `last_silence_mode_active` latch for a board its snapshot has not seen. Without that, an install booting **inside** its silence window reported a boundary the initial `check_and_send_active_page()` had already handled and forced one spurious extra update — #1817's `test_paused_board_does_not_force_an_update_every_second_during_silence` and `test_render_failure_...` caught it after the rebase. The snapshot is still recorded, so a board the drive path never touches (a disabled secondary) cannot re-trigger a forced update every second.

## Status

- Python: `5415 passed, 1 skipped` (full suite, in the CI image; `test_check_image_formats.py` excluded — it shells out to `git ls-files`, which cannot resolve a worktree `.git` file from inside the container). `ruff==0.16.4` check + format clean.
- Web: `121 passed` across the silence unit suites (`silence-schedule-per-board`, `silence-schedule-extended`, `silence-imminent-banner`, `active-page-display-silence`, `api-extended`); prettier + eslint clean on the touched files.
- Docs: cspell 0 issues, markdownlint 0 issues.
- E2E: the first CI run was `579 passed, 1 failed` — that one failure is the `/silence-status` layer bug documented above, now fixed.

### DIFF
--- a/docs/features/home-assistant-control.md
+++ b/docs/features/home-assistant-control.md
@@ -128,7 +128,7 @@ is never rewritten.
 |--------|------|-------------|
 | **Current Page** | Sensor | Name of the page currently displayed. *Includes JSON attributes (page_id, page_index).* |
 | **Board Message** | Sensor | Summary of what's currently on the board |
-| **Silence Mode** | Binary Sensor | Whether silence/quiet hours are active |
+| **Silence Mode** | Binary Sensor | Whether quiet hours are active on the **primary** board (silence is per board — see [Silence Schedule](./silence-schedule.md)) |
 | **Page Count** | Sensor | Number of pages configured *(state_class: measurement)* |
 
 ### Diagnostics (device health and info, grouped separately in HA)
--- a/docs/features/silence-schedule.md
+++ b/docs/features/silence-schedule.md
@@ -19,20 +19,33 @@ Split-flap displays can be noisy when updating. The silence schedule ensures you
 ### Via the Web UI
 
 1. Open the FiestaBoard Web UI at `http://localhost:4420`
-2. Go to **Settings**
+2. Go to **Settings** → **Behavior**
 3. Find the **Silence Schedule** section
 4. Set your **Start Time** and **End Time**
 5. Save your settings
 
-### Via Environment Variables
+Settings save automatically a second after you stop typing.
 
-You can also configure silence hours in your `.env` file:
+## Multiple Boards
 
-```bash
-# Silence Schedule
-SILENCE_SCHEDULE_START_TIME=22:00    # 10:00 PM
-SILENCE_SCHEDULE_END_TIME=07:00      # 7:00 AM
-```
+Silence settings are **per board**. If you run a Flagship in the living room
+and a Note on your desk, each keeps its own quiet hours, its own silence mode,
+and its own silence page — the Flagship can go quiet at 10 PM while the Note
+stays live, and each can show a page sized for its own display.
+
+Pick the board you want with the board selector in the sidebar, then edit the
+Silence Schedule card; you are editing that board's settings.
+
+A few things follow from this:
+
+- **A board you have not configured inherits the install-wide schedule.** That
+  is the schedule you had before you ever touched a second board, so upgrading
+  changes nothing: every existing board keeps the quiet hours it already had.
+  A board you add later starts out quiet during the same window rather than
+  flipping at 3 AM.
+- **The silence page picker only offers pages that fit that board.** A 22×6
+  Flagship page cannot be chosen for a 15×3 Note.
+- **Single-board installs see no difference.** There is just the one schedule.
 
 ## How It Works
 
@@ -45,11 +58,14 @@ how to handle the board based on the configured **silence mode**:
 | **Leave board unchanged** | The board is left exactly as it was — no further updates are sent until silence ends. |
 | **Show a specific page** | A page you choose is rendered once when silence begins and frozen on the board. Template variables are not refreshed until silence ends. |
 
+The mode is chosen per board, and so is the page: the silence page is always
+rendered at the size of the board it is being sent to.
+
 Regardless of which mode you pick:
 
-- The display service stops sending refresh updates while silenced
+- The display service stops sending refresh updates to that board while it is silenced
 - The web UI and API continue to function normally
-- Scheduled pages are skipped during silence hours
+- Scheduled pages are skipped during that board's silence hours
 
 When the silence window ends:
 
@@ -62,19 +78,17 @@ The silence schedule uses your configured timezone (`TIMEZONE` in `.env`). Make 
 
 ## Example Configurations
 
-### Nighttime Quiet Hours
+| Goal | Start Time | End Time |
+| --- | --- | --- |
+| Nighttime quiet hours | `22:00` (10 PM) | `07:00` (7 AM) |
+| Quiet outside office hours | `18:00` (6 PM) | `08:00` (8 AM) |
 
-```bash
-SILENCE_SCHEDULE_START_TIME=22:00    # Board goes quiet at 10 PM
-SILENCE_SCHEDULE_END_TIME=07:00      # Board resumes at 7 AM
-```
-
-### Office Hours Only
-
-```bash
-SILENCE_SCHEDULE_START_TIME=18:00    # Quiet after 6 PM
-SILENCE_SCHEDULE_END_TIME=08:00      # Resume at 8 AM
-```
+:::caution
+`SILENCE_SCHEDULE_START_TIME` and `SILENCE_SCHEDULE_END_TIME` appear in
+`env.example`, but the display service does not read them — they are written
+to a config key nothing consumes. Set your quiet hours in the Web UI (or via
+`PUT /settings/silence-schedule`) instead.
+:::
 
 ## Next Steps
 
--- a/docs/features/silence-schedule.md
+++ b/docs/features/silence-schedule.md
@@ -53,7 +53,8 @@ A few things follow from this:
   say it should be asleep.
 - **The Home Assistant `silence_mode` sensor reports the primary board.** Like
   the other entities in that payload (active page, transition style), it
-  describes the primary board rather than the install as a whole.
+  describes the primary board rather than the install as a whole. The same is
+  true of `GET /silence-status` when you do not name a board.
 
 ## How It Works
 
--- a/docs/features/silence-schedule.md
+++ b/docs/features/silence-schedule.md
@@ -45,7 +45,15 @@ A few things follow from this:
   flipping at 3 AM.
 - **The silence page picker only offers pages that fit that board.** A 22×6
   Flagship page cannot be chosen for a 15×3 Note.
-- **Single-board installs see no difference.** There is just the one schedule.
+- **Single-board installs see no difference.** There is just the one board, so
+  its schedule is the schedule.
+- **Manual sends respect the target board's window.** Sending a message or a
+  page from the Web UI, the API, or Home Assistant is blocked while *that*
+  board is silenced, so a 2 AM send cannot wake a board whose own quiet hours
+  say it should be asleep.
+- **The Home Assistant `silence_mode` sensor reports the primary board.** Like
+  the other entities in that payload (active page, transition style), it
+  describes the primary board rather than the install as a whole.
 
 ## How It Works
 
--- a/docs/reference/api-endpoints.md
+++ b/docs/reference/api-endpoints.md
@@ -152,7 +152,7 @@ pickers filter to the boards they're compatible with (see
 | `end_time` | string | Window end, UTC ISO |
 | `mode` | string | `"freeze"` (default), `"indicator"`, or `"page"` |
 | `page_id` | string \| null | Page shown when `mode` is `"page"`. Required in that mode, and validated against the target board's size |
-| `indicator_text` | string \| null | Text shown when `mode` is `"indicator"`. Uppercased; defaults to `SNOOZING` |
+| `indicator_text` | string \| null | Text shown when `mode` is `"indicator"`. Trimmed and forced to uppercase; defaults to `SNOOZING` |
 | `indicator_position` | string \| null | `"center"` (default), `"top-left"`, `"top-right"`, `"bottom-left"`, `"bottom-right"` |
 | `board_id` | string | Optional. Unknown board → `404`. **Note the exception below** |
 
--- a/docs/reference/api-endpoints.md
+++ b/docs/reference/api-endpoints.md
@@ -156,13 +156,23 @@ pickers filter to the boards they're compatible with (see
 | `indicator_position` | string \| null | `"center"` (default), `"top-left"`, `"top-right"`, `"bottom-left"`, `"bottom-right"` |
 | `board_id` | string | Optional. Unknown board → `404`. **Note the exception below** |
 
-Silence is the one per-board setting where omitting `board_id` does *not*
-mean "the primary board": it writes the **install-wide** schedule, which every
-board without its own override resolves to. That way an existing install keeps
-one set of quiet hours until you deliberately give a board its own, and a board
-you add later inherits the install's quiet hours rather than staying loud.
-`GET /silence-status?board_id=` returns the resolved values for that board and
-echoes `board_id` back.
+Reads and writes here use `board_id` differently, on purpose:
+
+- **`PUT /settings/silence-schedule` without `board_id`** writes the
+  **install-wide** schedule — the default every board without its own override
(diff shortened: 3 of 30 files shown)


jeffredodd:

```
```

Issue #1791, opened 2026-08-31
Board updates not working, only "Blank board" and "Show Debug Info" work


mountainsandcode:
**Describe the bug**
Thanks for building this app!

Any update I make to the board (e.g., switch pages) is supposedly successful, but the board never updates. 

The story is different when I use "Blank board" and "Show debug info on board" buttons in advanced settings - these always work instantly.

I also seem to have contradictory information whether the connection is actually happening via Local API or Cloud API - leading me to believe it might be some kind of API access-related issue?

**To Reproduce**
Try to switch any page, or create another update of the board.

**Expected behavior**
Board updates correctly or alternatively error message why board updates are not successful is shown (or why connection isn't working correctly).

I had it working with an earlier version of Fiestaboard 8.27, but since the board position has changed, so I can't exclude the possibility that it is related to this somehow (but the app does say the API is connected fine).

**Screenshots / Logs**

<img width="983" height="334" alt="Image" src="https://github.com/user-attachments/assets/d8dfb5b8-3508-4fd3-9b09-d4a3b6275752" />

<img width="957" height="208" alt="Image" src="https://github.com/user-attachments/assets/199069e4-3492-4758-94a9-0cf656bcc437" />

<img width="625" height="244" alt="Image" src="https://github.com/user-attachments/assets/f99032d7-9558-4217-848f-d78445cb37f2" />

**Environment**
- OS: [e.g. macOS 14, Ubuntu 22.04, Windows 11] Synology
- Docker version: [e.g. 24.0.7] DSM 7.3
- FiestaBoard version: [e.g. 1.32.59] 8.33.9
- Board connection mode: [local / cloud] Under System Information (Settings, Advanced), it says CLOUD API, but when I check the Board setting, it says it is connected successfully via Local API.
- Hardware: [e.g. Mac, Raspberry Pi 4, etc.] Synology DSM

**Additional context**

mountainsandcode:
Hmm, switching to a separate installation on another device resolved this issue - seems like a network issue

PR #1802, opened 2026-09-03
fix(debug): report real connection mode and surface silent send failures

jeffredodd:
Debug endpoints read wizard-era `config.json` rather than the `boards[]` store, so they disagreed with both Settings → Boards and the actual send path: System Info could say `CLOUD API` while Board settings said Local API. Send failures were also reported as success.

Issue #1791 is already closed, so this does not close anything — it lands the fix on its own terms.

## What changed

- `/debug/system-info`, `/debug/info`, `/debug/network-diagnostics` derive connection mode / board IP / credentials from `boards[]`, with live-client and legacy-`Config` fallbacks
- `PUT /settings/active-page` reports failures via an `error` field and `sent_to_board=false` instead of silently succeeding
- `/refresh` and `/force-refresh` propagate the send result and return 500 with the recorded reason
- `check_and_send_for_board` records `last_send_error` so endpoints can distinguish real failures from unchanged-content skips
- web: toast when the page switched but the board send failed (`toastSwitchNotSent`, all 14 locales)

## Review follow-ups

**Reverted the `_apply_env_overrides` change entirely.** Letting an env var override a value that merely *equals* the schema default is unsound: the code cannot distinguish "user deliberately chose the default" from "never configured". `_apply_env_overrides()` runs on every `ConfigManager.__init__` **and every `reload()`**, and `docker-compose.yml` loads the user's whole `.env`. Because `PUT /config/board` calls `Config.reload()` inline, a user who picked Local API in Settings — or in the setup wizard — had it silently reverted to `BOARD_API_MODE` within the same request. The same held for `general.timezone`, which feeds `time_service` and `schedules/sun_times`, so schedule rotations and sunrise/sunset would fire at the wrong time.

`src/config_manager.py` is now byte-identical to `main`. Two tests that encoded the unsound contract are gone; three that pin the correct one are new:

- a stored value equal to the schema default survives `_apply_env_overrides`
- it also survives a `set_board` + `reload()` round-trip (the `PUT /config/board` path)
- the same for `general.timezone`

The pre-existing `test_apply_env_overrides_preserves_user_customized_api_mode` only proved a *non-default* value survives, which is why the change looked correct.

**Closed the send-status race.** `rt.last_send_error` is shared with the engine thread, which could set or clear it between `check_and_send_for_board` returning and the endpoint reading it — giving a false success or a spurious 500. `check_and_send_for_board_with_status` / `check_and_send_active_page_with_status` now capture the reason for the calling thread only (a `threading.local`), and the endpoints take the reason from the call's return value rather than reading the runtime afterwards.

**`/refresh` no longer hides a failing secondary board.** With `board_id=None` the pass drives every board, but only the primary runtime's error was read, so a failing secondary returned success. The aggregate status now reports the first failure across all boards, prefixed with its board id.

## Tests

`tests/test_per_board_engine.py` gains a `TestSendStatusReporting` class covering the aggregate across boards, the unprefixed primary reason, and a concurrent write from another thread being ignored. `tests/test_api_coverage.py` pins that a stale `last_send_error` cannot turn a clean pass into a 500.

Verified non-vacuous: expressed against the pre-fix read strategy (drive, then read `service.get_last_send_error()`), both defects reproduce —

```
E   AssertionError: assert None == 'board b2: secondary board render failed'
E   AssertionError: engine thread's error leaked into this call: 'engine tick failure'
```

and the reverted env-override tests failed against the branch code before the revert:

```
E   AssertionError: assert 'cloud' == 'local'      # board.api_mode after reload()
E   AssertionError: assert 'Europe/Berlin' == 'America/Los_Angeles'
```

592 Python tests pass across the engine / API / config / settings suites; `ruff 0.16.4` check + format clean; web prettier clean.

### DIFF
--- a/src/api_server.py
+++ b/src/api_server.py
@@ -3209,24 +3209,29 @@ async def stop_service():
     return {"status": "stopped", "message": "Service stopped successfully"}
 
 
-def _service_send_error(service, board_id: str | None = None) -> str | None:
-    """The failure reason recorded by the service's last send attempt, or None.
+def _send_with_status(service, method: str, fallback: str, *args, **kwargs) -> tuple[bool, str | None]:
+    """Run a ``check_and_send_*`` pass and return ``(sent, failure reason)``.
 
     ``check_and_send_*`` swallow exceptions and return a bool that conflates
-    "failed" with benign skips; the recorded ``last_send_error`` is what lets
-    endpoints report real failures (issue #1791). Only a non-empty ``str``
-    counts — Mock services from older test fixtures return Mock attributes,
-    which must not read as failures (same convention as ``_board_is_paused``).
+    "failed" with benign skips, so endpoints reported silent failures as
+    success (issue #1791). The ``*_with_status`` wrappers capture the reason
+    for *this* call in a thread-local, which is why the reason must come back
+    from the call rather than be read off the runtime afterwards — the engine
+    thread rewrites ``last_send_error`` on its own cadence.
+
+    Falls back to the plain method (and no reason) when the service does not
+    expose the wrapper, so Mock services from older test fixtures still work.
+    Only a non-empty ``str`` counts as a reason, for the same Mock reason
+    (same convention as ``_board_is_paused``).
     """
-    getter = getattr(service, "get_last_send_error", None)
-    if not callable(getter):
-        return None
-    try:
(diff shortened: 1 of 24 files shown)


jeffredodd:

```
```

PR #1803, opened 2026-09-03
feat(settings): allow renaming boards

### LINKED ISSUE #1792: Allow renaming of boards / definition of board name
It would be great if boards can be renamed so they are not just called "My board" or "My board 2"

User-definable board name.

**Additional context**

jeffredodd:
Closes #1792.

Adds an editable Name field to each board card in Settings → Boards, saved through the existing `PUT /settings/board` round-trip. Trimmed on blur; clearing it saves `""` and the backend restores its default (`My Board`).

The setup wizard no longer hardcodes `name: "My Board"` — it omits the field so the backend default applies, keeping display-name policy out of the wizard.

Adds `boardNameLabel` to all 14 locales.

## Review follow-ups

**The name input is now controlled.** It was `defaultValue`, so the server-normalized name never landed in the field: clearing the name flipped the card header to "My Board" while the input still showed `""`.

```
after save, input           = ""
after collapse+reopen, input = "My Board"
```

Because the field stayed stale, `trimmed !== board.name` was permanently true, so **every subsequent focus/blur fired another identical PUT** — each one running `_reinitialize_board_clients()` and rewriting `config.json` (observed `put count = 2`). A `BoardNameField` component now owns a draft and re-syncs from the query data whenever the stored name changes underneath it, which fixes both.

**Whitespace-only names no longer persist.** `BoardInstance.__post_init__` only had `if not self.name: self.name = "My Board"`, and `"   "` is truthy — so a name of spaces was stored verbatim and rendered as a blank sidebar row. It now strips, caps at `MAX_BOARD_NAME_LENGTH` (64, since the name is rendered in the sidebar, board selector and card headers), then falls back to the default. The input carries a matching `maxLength` so the field never shows more than what will actually be stored.

**A rename breaks name-based MQTT automations.** `CommandHandler._resolve_board` matches by id **then by display name**, so shipping a rename UI silently breaks any Home Assistant automation publishing `{"board": "Kitchen", ...}`. FiestaBoard cannot rewrite payloads that live in the user's HA config, so name resolution now logs a warning naming the stable id, and `docs/features/home-assistant-control.md` gains a troubleshooting entry. HA discovery topics are id-scoped, so entities are never orphaned — only name-based `board` refs in user automations.

## Test fixes

**`blurring without changing the name fires no PUT` was vacuous.** `put.body` is assigned inside an *async* MSW handler and was asserted synchronously, so it read `null` whether or not a request fired. Proven: deleting the production guard left all three new tests green.

```
Test Files  1 passed (1)
     Tests  33 passed (33)      # guard deleted
```

It now awaits and pins a request *count*. With the guard deleted again, it goes red:

```
 ✓ ... shows the current name and saves an edited, trimmed name on blur
 × ... blurring without changing the name fires no PUT
 ✓ ... clearing the name saves an empty string and shows the restored default
 × ... re-blurring after a clear does not fire a second identical PUT
 Test Files  1 failed (1)
      Tests  2 failed | 32 passed (34)
```

**The MSW fixture echoed `name` back verbatim** instead of applying the backend's `name or "My Board"`, so the clear/reset round-trip this PR is *about* was never exercised. It now mirrors `BoardInstance.__post_init__` (strip → cap → default). Against the uncontrolled input the two new tests fail exactly where expected:

```
AssertionError: expected '   ' to be 'My Board' // Object.is equality
Expected: "My Board"
Received: "   "
```

**`test_set_boards_roundtrips_renamed_board` / `test_set_boards_empty_name_restores_default` both passed unchanged against `main`** (the original PR changed no backend code), so they were not evidence of anything. They are joined by whitespace-only, trim, and cap cases that do fail against `main` — 4 red before the `devices.py` change:

```
E   AssertionError: assert '   ' == 'My Board'
E   AssertionError: assert '  Kitchen Board  ' == 'Kitchen Board'
E   AssertionError: assert 200 == 64
E   AssertionError: assert '  Garage  ' == 'Garage'
```

## Verification

- Python: 517 passed (devices / settings / mqtt / api / board suites); `ruff 0.16.4` check + format clean
- Web: 134 files, 1554 passed, 7 skipped; prettier, eslint and `tsc --noEmit` clean

### DIFF
--- a/docs/features/home-assistant-control.md
+++ b/docs/features/home-assistant-control.md
@@ -444,7 +444,7 @@ breaks every automation that targets it by the old name — FiestaBoard cannot
 rewrite payloads that live in your Home Assistant config. When this happens
 the log records:
 
-```
+```text
 MQTT: resolved board by name 'Kitchen'; renaming this board will break it. Target the board id 'b1' instead.
 ```
 
--- a/docs/features/home-assistant-control.md
+++ b/docs/features/home-assistant-control.md
@@ -430,6 +430,28 @@ FiestaBoard properly categorizes entities using Home Assistant's `entity_categor
 - Check that HA's MQTT integration is configured with the correct broker credentials
 - Check FiestaBoard logs for command processing errors
 
+### A Command Stopped Working After Renaming a Board
+
+Command payloads can target a specific board with a `board` field, which
+accepts either the board's id or its display name:
+
+```json
+{ "board": "Kitchen", "page_id": "weather" }
+```
+
+Display names are editable in **Settings → Boards**, so renaming a board
+breaks every automation that targets it by the old name — FiestaBoard cannot
+rewrite payloads that live in your Home Assistant config. When this happens
+the log records:
+
+```
+MQTT: resolved board by name 'Kitchen'; renaming this board will break it. Target the board id 'b1' instead.
+```
+
+Target the board **id** instead. Ids never change, and they are what the
+discovery topics already use, so renaming a board never orphans its HA
+entities — only name-based `board` refs in your own automations.
+
 ## Branding
 
 ### Entity Icons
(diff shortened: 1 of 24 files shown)


jeffredodd:

```
```

Issue #1744, opened 2026-08-24
MCP bearer token unenforced when auth mode is "disabled"; CORS wildcard + credentials


jeffredodd:
### Why
`AuthMiddleware` returns early for `auth_mode() == "disabled"`
(`src/auth/middleware.py:111`) before the MCP bearer branch — so
`FIESTABOARD_MCP_TOKEN` provides false assurance and 28 mutating tools are open on
the LAN, with DNS-rebinding protection deliberately off (`src/mcp_server.py:1351`).
Separately `allow_origins=["*"]` + `allow_credentials=True`
(`src/api_server.py:891`) is an invalid combination covering the whole app
including the MCP mount.
### What
1. Enforce the MCP bearer check whenever a token is configured, regardless of auth
   mode.
2. Fix CORS: no wildcard with credentials (echo specific origins or drop
   credentials).
### Acceptance
Auth-disabled + token set → 401 without bearer on /api/mcp; CORS config valid.

_Part of #1730 (Phase 1)._

PR #1851, opened 2026-09-06
chore: remove unrouted board-config compat handlers and their coverage-inflating tests

jeffredodd:
Closes #1747. Part of #1730 (Phase 1, last open item) — first PR of the backend-rework Stack 0.

## Why

`get_board_config_compat` / `update_board_config_compat` in `src/api_server.py` have no route decorator and the file has no dynamic route registration (`add_api_route` grep: zero hits) — they are unreachable over HTTP. Yet `tests/test_api_coverage_extended.py::TestDeprecatedCompat` imported the coroutines and called them directly with a `Mock` standing in for `Response`, manufacturing coverage for dead code: those tests would pass even if no route ever existed.

## What

Pure deletion:
- the two handlers (16 lines) in `src/api_server.py`
- the `TestDeprecatedCompat` class (43 lines) in `tests/test_api_coverage_extended.py`

A repo-wide grep for `board_config_compat` confirms these were the only references — no frontend, docs, or plugin code touches them. The real `/config/board` GET/PUT handlers are untouched.

## Zero-regression evidence

- Route-inventory golden test (#1731) passes **unchanged** — proving nothing routed was removed.
- `tests/test_route_inventory.py` + `tests/test_api_coverage_extended.py`: **69 passed** in the isolated test container.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01Par2E2Nh65PyDF1q9TARJP

Issue #857, opened 2026-06-02
Clicking Settings -> Integrations causes infinite page refresh


cordell25:
**Describe the bug**
Clicking on Settings -> Integrations sends the webpage into an infinite loop where it appears to be redirected back to itself over and over. Note, I am not using the 'secure' login feature. The only way out is to force close the browser tab.

**Environment**
- FiestaBoard version: [6.10.0]
- Board connection mode: [local]
- Hardware: [Raspberry Pi 3]

PR #1853, opened 2026-09-06
fix(auth): possession-gate MCP token management when auth is disabled

### LINKED ISSUE #1825: security: MCP token-management routes are unauthenticated when auth mode is disabled (_require_admin returns "anonymous")
## Summary

`_require_admin()` in `src/auth/routes.py` returns the sentinel string
`"anonymous"` instead of raising when `auth_mode() == "disabled"`. Every
MCP token-management route is gated by that helper, so on an install with
the browser login turned off those routes are **unauthenticated**.

This is pre-existing behaviour, not a regression from #1810 — but #1810 is
what makes it load-bearing. That PR closed #1744 by enforcing a configured
MCP bearer token on `/api/mcp/` in every auth mode. The data path is now
genuinely gated; the *management* path that can mint or destroy the
credential is not.

Net effect on the exact deployment class #1810 targets — auth off, MCP
token set via **Settings → Integrations**, LAN-reachable — anyone who can
reach the port can obtain a valid token, or revoke the stored one and
re-open all 28 mutating MCP tools.

## Why the sentinel exists

It is deliberate, and the docstring explains it:

> When auth is explicitly *disabled* (...) there is no admin concept and no
> session cookie ever gets issued — return a sentinel so MCP token
> management remains reachable. Without this the Settings → Integrations
> page hits a 401, the web client redirects to `/login`, the login page
> sees auth is disabled and bounces back, and the user is stuck in an
> infinite reload loop.

So a naive "just raise 401" fix reintroduces the redirect loop. Any fix
needs to keep **Settings → Integrations** usable on an auth-disabled
install while closing the network path — e.g. a same-origin/CSRF check, a
localhost-only carve-out, or having the UI treat 401 on this route as "not
available" rather than "log in".

## Reproduction

Auth disabled, token stored via the Settings page (`source == "stored"`):

```text
anon GET    /mcp/                   -> 401   (#1810 working as intended)
anon POST   /auth/mcp-token         -> 201 {"token":"xkIfsZl0eRJBSVPIZe9kiRQ9bWizrsWw-AAHGHiodIo"}
anon DELETE /auth/mcp-token         -> 200 {"status":"ok","username":null}
token source after revoke: none
anon GET    /mcp/ after revoke      -> 404   (past the auth layer, into the MCP mount — open)
```

The `404` on the last line is the request sailing straight past the auth
layer into the MCP mount, the same signature as the original #1744 failure
mode.

The env-pinned configuration is **not** affected — `mcp_token_source() ==
"env"` short-circuits both routes before any state change:

```text
--- with FIESTABOARD_MCP_TOKEN pinned ---
anon POST   /auth/mcp-token         -> 409
anon DELETE /auth/mcp-token         -> 409
```

## Affected routes

All three are gated only by `_require_admin`:

- `GET /auth/mcp-token` — leaks whether a token is configured and its source
- `POST /auth/mcp-token` — mints and returns a valid plaintext token
- `DELETE /auth/mcp-token` — revokes the stored token

Worth auditing whether anything else calls `_require_admin` and has the
same exposure.

## Not in scope for #1810

#1810 is a targeted request-entry fix and deliberately does not touch
`_require_admin`. It documents the limitation instead:
`docs/setup/authentication.md` and `docs/internal/setup/MCP_CLIENTS.md`
both now say that on an auth-disabled install a Settings-stored token is a
guard against accidents rather than a security boundary, and that
`FIESTABOARD_MCP_TOKEN` is the configuration that actually holds.

## Workaround (available today)

Pin the token with `FIESTABOARD_MCP_TOKEN` instead of generating it in the
UI, or enable the browser login.

Refs #1744, #1810.

jeffredodd:
Closes #1825. Backend-rework **Stack 0** (stacked on #1851; retarget to main as the stack merges).

## The vulnerability

`_require_admin()` returns the `"anonymous"` sentinel when auth mode is disabled, so on an auth-disabled, LAN-reachable install **anyone who can reach the port** could `POST /auth/mcp-token` (mint and read a valid token) or `DELETE` it (revoke, re-opening `/api/mcp` entirely) — the exact management-path asymmetry #1810 left open after gating the data path.

## The fix: possession-based management

New `_require_admin_or_mcp_token()` gate on all three `/auth/mcp-token` routes:
- enabled/undecided modes: unchanged (session auth)
- disabled + **no token configured**: open — first mint stays anonymous by design (with auth disabled the whole REST surface is open; gating the first mint protects nothing)
- disabled + token configured (stored **or** env): caller must present the *current* token as `Authorization: Bearer`; otherwise 401 + `WWW-Authenticate` challenge. Status included — it leaks the token's source.
- env-pinned installs: possession gate first (anon → 401), then the existing 409 pin on mutation

No login-redirect loop (the #857 regression this sentinel guarded against): the web client now treats 401 on this route as "locked", not "sign in" — a password input in Settings → Integrations unlocks management with the current token, and a rotate hands the UI the new token so management continues.

`_require_admin` audit: zero other callers repo-wide (grep evidence in the commit); every other auth route is deliberately public or re-verifies the password inline.

## Zero-regression evidence

**Fail-first (backend)** — the new tests reproduce the issue's exact repro before the fix:
```
tests/test_auth_mcp_token_routes.py .............FFFFFFF.
E   AssertionError: assert 201 == 401   (anon second POST)
E   assert 200 == 401                   (anon DELETE with stored token)
E   assert 200 == 401                   (anon GET with stored token)
========================= 7 failed, 14 passed =========================
```
The three tests that previously **pinned the vulnerability** are rewritten with in-place justification (repo precedent: the #1744 contract reversal in `test_auth_mcp_bearer.py`). No other assertion weakened.

**After:** auth suites + CORS + route-inventory golden: **110 passed**. Web: new locked-state tests **4/4** (non-vacuity proven by reverting the redirect-skip → 4/4 fail, and the held-token update → 1 fails on the exact assertion); full web suite **1611 passed / 0 failed**; `tsc`, eslint, prettier clean.

Docs (`docs/setup/authentication.md`, `docs/internal/setup/MCP_CLIENTS.md`) updated from "known hole + workaround" to the new contract.

### DIFF
--- a/docs/internal/setup/MCP_CLIENTS.md
+++ b/docs/internal/setup/MCP_CLIENTS.md
@@ -64,8 +64,15 @@ client config — it replaces `<YOUR_TOKEN>` in the examples below.
 > login-redirect loop. (Earlier releases left these routes fully anonymous
 > on such installs, so anyone who could reach the port could mint or
 > revoke the token — fixed in Fiestaboard/FiestaBoard#1825.)
-> `FIESTABOARD_MCP_TOKEN` is still the strongest configuration: while it
-> is set, the mutating routes refuse with `409` even for a caller
+> Even so, the possession gate is **not a boundary against an attacker
+> who can already reach the port**: on an install where the login is
+> merely *disabled by preference*, the preference itself is part of the
+> open API, so such an attacker can `POST /auth/preference` to enable the
+> login, register the first admin via `POST /auth/setup`, and manage the
+> token with that session. Treat the gate as protection against casual or
+> accidental mint-and-read and revocation.
+> `FIESTABOARD_MCP_TOKEN` is the configuration that actually holds: while
+> it is set, the mutating routes refuse with `409` even for a caller
 > presenting the token, so it cannot be rotated or revoked over the
 > network at all. Enabling the browser login gates management behind the
 > admin session as usual.
--- a/docs/internal/setup/MCP_CLIENTS.md
+++ b/docs/internal/setup/MCP_CLIENTS.md
@@ -52,22 +52,23 @@ client config — it replaces `<YOUR_TOKEN>` in the examples below.
 > configured, nothing changes: `/api/mcp/` follows whatever the auth mode
 > says.
 
-> **With the login disabled, pin the token with the env var.** The paragraph
-> above is about the `/api/mcp/` data path only. The *token management*
-> routes are not gated the same way: `_require_admin` in
-> `src/auth/routes.py` deliberately returns an `"anonymous"` sentinel when
-> `auth_mode()` is `disabled`, so that **Settings → Integrations** doesn't
-> bounce off a 401 into a login-redirect loop. The consequence is that on an
-> install with the login off, an unauthenticated caller who can reach the
-> port can `POST /auth/mcp-token` to mint a valid token, or
-> `DELETE /auth/mcp-token` to revoke the stored one and re-open `/api/mcp/`
-> outright. A token stored from the Settings page is therefore a guard
-> against accidental clients on such an install, **not** a boundary against
-> an attacker who can already reach FiestaBoard. Setting
-> `FIESTABOARD_MCP_TOKEN` is the configuration that holds: while it is set,
-> both management routes refuse with `409`, so the token cannot be rotated
-> or revoked over the network. Enabling the browser login also closes it.
-> Tracked in Fiestaboard/FiestaBoard#1825.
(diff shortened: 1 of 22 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/fix/1825-mcp-token-mgmt-auth origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

Issue #1826, opened 2026-09-04
Blocking I/O on the event loop: /refresh, /force-refresh, and page preview still freeze the API


jeffredodd:
Follow-up to #1750. PR #1809 converted the plugin install/update family (and the hourly auto-update loop) to `asyncio.to_thread`, but it deliberately scoped itself to those paths. The other three families named in #1750 still run their blocking work inline on the single event loop, so this issue tracks the remainder and #1750 should not be closed until it is done.

### What still blocks

| Path | Location | Blocking work on the loop |
| --- | --- | --- |
| `POST /refresh` | `src/api_server.py:3109` (`refresh_display`) | full render + 10 s board POST + up to 120 s transition |
| `POST /force-refresh` | `src/api_server.py:8776` (`force_refresh`) | same render + board POST path |
| page preview | `src/api_server.py:7804` (`preview_page`) -> `:7828` | `page_service.preview_page()` runs the whole plugin fan-out inline |

While auditing for this issue the preview surface turned out to be wider than #1750 named. Every one of these is an `async def` calling straight into the synchronous page service:

- `src/api_server.py:6047` — `set_active_page`
- `src/api_server.py:7619` — `get_current_display`
- `src/api_server.py:7828` — `preview_page`
- `src/api_server.py:7873` — `preview_pages_batch` (N plugin fan-outs in one request)
- `src/api_server.py:7990` — `send_page`

`preview_pages_batch` is the sharpest of them: one request pays the sum of every page's plugin fetches, with the loop held for the whole total.

### Why

Same mechanism as #1750: FiestaBoard is a single-process ASGI app with one event loop. For the duration of any of these the board stops being driven, every page in the web UI hangs, and `/health` does not answer.

### What

Wrap the blocking sections in `asyncio.to_thread`, or make the handlers plain `def` and let Starlette threadpool them. The pattern is already established across `src/api_server.py` and in PR #1809.

Note the ordering constraint that install did not have: several of these paths mutate shared display/page state, so moving them off the loop interacts with the registry/state locking gap tracked separately (see the companion issue filed alongside this one). Take that into account rather than assuming the install conversion transfers unchanged.

### Acceptance

Extend `tests/test_plugin_install_event_loop.py` (or a sibling in the same watchdog style) so `GET /health` is served while a slow stubbed refresh / preview is in flight, with health latency well under the existing 0.5 s budget. Confirm each new test fails before its fix.

Refs #1750, #1809.

[...]

PR #1854, opened 2026-09-06
fix(plugins): lock registry/loader shared state; serialize per-directory git ops

### LINKED ISSUE #1828: Plugin registry state is unlocked: off-loop install races every API request
Follow-up to #1750, surfaced by review of PR #1809.

### The trade PR #1809 made

Moving plugin install/update to `asyncio.to_thread` fixed the API freeze, but it gave up serialization the event loop was providing for free. While install ran inline, nothing else in the process could touch the registry — not because anything was locked, but because there was only one thread of control. Off-loop, that guarantee is gone.

`PluginRegistry` / `PluginLoader` mutable state is entirely unlocked: `_plugins`, `_loaded_plugins`, `_plugin_classes`, and `sys.modules`.

### Concrete failure

`GET /plugins` calls `registry.list_plugins()` (`src/plugins/registry.py:1096`), which iterates `self._plugins`. Concurrently, `install_from_registry` does `self._plugins[plugin_id] = plugin` (`src/plugins/registry.py:1415`; also `:315`, `:1164`, `:1457`). Interleaved, that is:

```
RuntimeError: dictionary changed size during iteration
```

Second failure mode: two concurrent `POST /plugins/{id}/update` calls can now both be inside `clone_or_update_repo` (`src/plugins/sources.py:310`) running `git fetch` / `git reset --hard` against the *same* directory. That surfaces to the user as an opaque `index.lock` 500.

### Scope — this is a widened window, not a new bug

The display thread already raced the registry this way before #1809, so the bug class predates it. What changed is the width: previously the only racer was one background thread, now it is every API request that touches the registry while an install or update is in flight. #1809 documents this deliberately rather than attempting the locking refactor inline, because a correct fix is a refactor and not a two-line change.

### What

Introduce a registry mutex — a `threading.Lock` (or `RLock`) guarding registry mutation and iteration, with an eye to:

- readers (`list_plugins`, `get_plugin`, `get_load_errors`) taking it too, or the mutations swapping in a new dict rather than mutating in place
- a per-plugin-directory lock (or a single global update lock) so two `clone_or_update_repo` calls cannot share a working tree
- **not** disturbing the deliberate executor pattern in `src/plugins/registry.py` — the `shutdown(wait=False, cancel_futures=True)` there is intentional and must not be tidied into a `with` block
- holding the lock only around state transitions, not around the 120 s `git fetch`, or the freeze comes straight back in a new form

### Acceptance

A test that hammers `GET /plugins` against a concurrent install and does not raise `RuntimeError: dictionary changed size during iteration`; and a test that two overlapping updates of the same plugin serialize instead of both entering the git working tree. Both must fail before the fix.

Refs #1750, #1809.

jeffredodd:
Closes #1828. Backend-rework **Stack 0**, stacked on #1853 (chain: #1851 → #1853 → this). Prerequisite for moving the remaining blocking I/O off the event loop (#1826).

## Why

PR #1809 moved plugin install/update to `asyncio.to_thread`, which traded away the serialization the single event loop provided for free. `PluginRegistry`/`PluginLoader` mutable state was entirely unlocked (`threading` wasn't even imported in registry.py): `GET /plugins` iterating `_plugins` during a concurrent install raises `RuntimeError: dictionary changed size during iteration`, and two overlapping updates of the same plugin could both run `git fetch`/`git reset --hard` in one working tree (opaque `index.lock` 500s).

## Design

- **One shared `threading.RLock`** created by `PluginRegistry`, passed into its `PluginLoader` — a single lock across both objects because the registry reaches into loader state on the install paths; no ordering deadlocks possible.
- **Snapshot-under-lock iteration** at every iteration site (list_plugins, enabled_plugins, trigger_plugins, variable maps, …); mutations guarded end-to-end.
- **The deadlock rule**: `build_template_context` snapshots `enabled_plugins` under the lock and *releases before* the fetch fan-out — workers re-acquire briefly inside `fetch_plugin_data`. The deliberate `shutdown(wait=False, cancel_futures=True)` executor pattern is untouched.
- **Never held across git**: install paths do the 120 s git work *before* taking the registry lock; the state swap alone is locked.
- **Per-plugin-directory lock** in `sources.py` around `clone_or_update_repo`'s git/rmtree body — same-plugin ops serialize, different plugins stay parallel; the CodeQL path-injection barrier block is wrapped, not restructured.
- New public `registry.clear_update_status()` replaces api_server's three direct `_update_status.pop()` reaches; two extra unlocked iteration sites found while sweeping (api_server transition listing, template engine `_manifests`) hardened the same way.

## Zero-regression evidence

**Fail-first** (first race-test draft was vacuous — writer finished before the reader overlapped; rebuilt with sustained churn + `sys.setswitchinterval(1e-6)`, then deterministic 6/6 pre-fix failures):
```
E   AssertionError: concurrent list_plugins/mutation raised: [RuntimeError('dictionary changed size during iteration')]
E   AssertionError: same-directory git ops overlapped (max concurrency 2)
```
Post-fix: new suite **5/5** (incl. an over-serialization guard proving different plugins still clone in parallel), stable across repeated runs.

**Regression**: 1566 passed across every plugin/registry/loader/engine/template/transition suite + `test_plugin_install_event_loop` + `test_plugin_lifecycle_round_trip` + route-inventory golden. The only 2 failures are the pre-existing `generative_dashboard` seed gap, fixed by #1852 (verified pre-existing via stash on the unmodified tree). Two `test_settings_plugins.py` assertions were re-wired to the new `clear_update_status` contract without weakening (the mock's side_effect still performs the pop so the same ids are verified).

### DIFF
--- a/src/api_server.py
+++ b/src/api_server.py
@@ -2137,7 +2137,7 @@ async def _auto_apply_plugin_updates(registry: Any, plugin_ids: list) -> None:
             failed.append(plugin_id)
             continue
 
-        registry._update_status.pop(plugin_id, None)
+        registry.clear_update_status(plugin_id)
         updated.append(plugin_id)
 
     if updated:
@@ -5727,7 +5727,9 @@ async def list_transition_plugins():
 
     registry = get_plugin_registry()
     out = []
-    for plugin_id, plugin in registry._plugins.items():  # noqa: SLF001 - registry is in-process
+    # registry.plugins copies under the registry lock, so a concurrent
+    # install/uninstall can't mutate the dict mid-iteration (#1828).
+    for plugin_id, plugin in registry.plugins.items():
         if not isinstance(plugin, TransitionPluginBase):
             continue
         manifest = registry.get_manifest(plugin_id)
@@ -10751,7 +10753,7 @@ async def update_plugin(plugin_id: str):
         detail = "; ".join(errors) if errors else "Plugin failed to reload after update."
         raise HTTPException(status_code=500, detail=detail)
 
-    registry._update_status.pop(plugin_id, None)
+    registry.clear_update_status(plugin_id)
 
     return {
         "status": "success",
@@ -10823,7 +10825,7 @@ async def apply_all_plugin_updates():
             failed[plugin_id] = "; ".join(errors) if errors else "Reload failed."
             continue
 
-        registry._update_status.pop(plugin_id, None)
+        registry.clear_update_status(plugin_id)
         updated.append(plugin_id)
         logger.info("Bulk update: applied update for plugin '%s'", plugin_id)
 
--- a/src/plugins/loader.py
+++ b/src/plugins/loader.py
@@ -649,11 +649,12 @@ class PluginLoader:
             Reloaded plugin instance, or None if failed
         """
         with self._lock:
-            # Unload if loaded
+            # Unload if loaded.  cleanup() is plugin-authored and unbounded,
+            # so it is retired to a daemon thread instead of running under
+            # the shared registry/loader lock (#1854).
             if plugin_id in self._loaded_plugins:
-                old_plugin, _ = self._loaded_plugins[plugin_id]
-                old_plugin.cleanup()
-                del self._loaded_plugins[plugin_id]
+                old_plugin, _ = self._loaded_plugins.pop(plugin_id)
+                self._retire_instance(plugin_id, old_plugin)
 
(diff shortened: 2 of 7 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/fix/1828-registry-locks origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1855, opened 2026-09-06
feat(tests): one data-dir seam + autouse isolation — the suite can no longer touch real data/

### LINKED ISSUE #1762: Global test isolation: stop the suite writing the developer's real data/
### Why
`tests/conftest.py` resets nothing; most tests read/write the real `data/` dir.
Nine files re-implement the same singleton-reset fixture. This is the cause of the
process-scoped-tmp workaround documented in `atomic_io.py:10-13` and of the known
full-suite pollution failures.
### What
One autouse conftest fixture: redirect the data dir to `tmp_path`, reset the four
singletons (ConfigManager, settings/page/display services); delete the nine local
copies.
### Acceptance
Full suite leaves `data/` untouched (checksum before/after); pollution-class
failures gone.

_Part of #1730 (Phase 3)._

jeffredodd:
Closes #1762. **Track A** base of the backend rework (stacked on #1851; will retarget as Stack 0 merges). First item of the platform-spine track because every later change needs a green suite that actually means green.

## Why

`tests/conftest.py` reset nothing; most tests read/wrote the developer's real `data/`. Nine files re-implemented the same singleton-reset fixture, and the app had **no data-directory seam at all** — eleven independent `Path(__file__)`-relative resolutions, several of which `mkdir` the real `data/` at construction. This is the root of the known full-suite pollution failures and the process-scoped-tmp workaround documented in `atomic_io.py`.

## What

- **The seam**: new `src/paths.py::get_data_dir()` — `FIESTABOARD_DATA_DIR` env var, else `<repo>/data`; read fresh on every call. All eleven resolution sites rewired (config, settings, pages, schedules, collections, panels, backup, auth, secrets, external-plugins dir, and api_server's two cwd-relative import-time constants → lazy accessors). Explicit-path constructor kwargs untouched.
- **The fixture**: autouse `_isolated_data_dir` in conftest — env var → `tmp_path`, singletons dropped on both sides in the registry-before-engine order the repo already documents as a hazard.
- **The guard**: CI now checksums `data/` before the platform-test step and fails on any write after it. New `tests/test_data_dir_isolation.py` asserts every store's *default* path resolves under the tmp dir.
- Local reset fixtures simplified in 8 files (env-specific parts kept); `test_plugin_lifecycle_round_trip.py`'s leak-guard detector deliberately kept.

## Zero-regression evidence

- **Fail-first**: the 9 isolation tests written before the seam — `9 failed` (e.g. `AuthService default resolved into the repo data/: /app/data/auth.json`); all pass after.
- **Full suite, baseline vs after** (isolated container, `-n auto`): `3 failed, 5606 passed` → **`1 failed, 5617 passed`** — a strict subset; the two config-manager **pollution failures vanished**, and the sole remainder is the known worktree-environment `test_check_image_formats` failure (passes in CI).
- One test retargeted to a **stricter** contract: `test_backup_import_restores_every_data_file_byte_for_byte` previously tolerated a migration-flag delta that only existed because of cross-test registry leakage; with per-test-fresh registries the restore is a true byte-for-byte no-op, asserted as exactly that with the premise pinned.


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1762-test-isolation origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

Issue #1850, opened 2026-09-06
Umbrella: engine redesign — change-driven tick, per-board send workers, demand-driven fetch


jeffredodd:
Part of the backend rework (spec approved 2026-09-05; extends epic #1730 Phase 2).

One coherent redesign of the display engine instead of five patches. Target
pipeline: inputs → context → resolution → render → per-board send queues, each
stage short-circuiting when its inputs are unchanged.

Landing order (stacked PRs):
1. **Equivalence harness** — fake-clock scenario corpus whose golden output is
   the sequence of board sends produced by the CURRENT engine; the new engine
   must reproduce it exactly except explicitly-allowlisted diffs, each of
   which is itself asserted. Gate for everything below.
2. Change-driven tick (#1752): one context per tick per board size; render
   short-circuits when inputs unchanged; silence window cached, invalidated on
   config write.
3. Demand-driven fetch (#1751): fan-out only to referenced + trigger-capable
   plugins; one persistent bounded executor.
4. Per-board send workers (#1755): tick enqueues; transitions block only
   their own board.
5. Board client policy (#1754): retry/backoff, universal min-send-interval,
   per-board throttle lock, short LAN connect timeout.
6. Board recovery (#1827): backoff retry of `board_init_errors` on the
   existing poll thread, swap synchronized with send workers.
7. Triggers integrated as a first-class input stage (per the #1767 decision:
   keep the plugin contract, persist dismissals on the storage kernel #1848).

Absorbs/extends: #1751, #1752, #1754, #1755, #1827; resolves #1767's
disposition. Depends on #1848 for the dismissal store only.

jeffredodd:
All seven landing-order items are on `next`; the last deviation (trigger dismissals writing through `atomic_io` instead of `JsonStore`) closed in #2018, which also adds the golden and concurrent-writer coverage the other stores have. Closing; ships with 9.0.

Issue #1734, opened 2026-08-24
E2E: display-loop timing under a fake clock — schedule boundary + collection rotation


jeffredodd:
### Why
`DisplayService.run()` (tick cadence, sleeps, boundary detection) is exercised only
for crash resilience; every hardware E2E is a single forced tick. No test crosses a
real schedule time boundary, and nothing asserts a rotating collection actually
advances what's displayed. Phase 2 rewrites this loop.
### What
pytest driving `run()` with an injected/fake time service and a cancel-after-N-ticks
sentinel:
1. Page switches when the clock crosses a schedule boundary.
2. A time-mode collection advances to its next page after `interval_seconds`.
3. Tick work happens at `polling_interval` cadence, not 1 Hz.
### Acceptance
Assertions on what was sent to a stub client per simulated timeline; no real sleeps
(suite stays fast).

_Part of #1730 (Phase 0)._

PR #1856, opened 2026-09-06
test(engine): equivalence harness — golden send-sequences pinning the current engine

jeffredodd:
Part of #1850 (engine-redesign umbrella). **Track B** base of the backend rework (stack #1875).

## What

A fake-clock scenario corpus whose golden output is the **sequence of board sends** `(sim-time, board, rows)` produced by the *current* engine, committed as `tests/golden/engine/*.json`. The engine redesign PRs stacked above must reproduce these sequences exactly; intentional diffs are explicit, asserted allowlist entries. Goldens deliberately do NOT encode tick counts or wake cadence (the #1734 discipline).

Nine scenarios, all driven through the real `DisplayService.run()` loop: static-page dedupe, changing-page cadence, schedule boundary, time-mode collection rotation, silence window, **two boards** (previously untested under `run()`), trigger lifecycle through the **real TriggerService**, board send failure + recovery, temporary inline override. Re-record with `RECORD_ENGINE_GOLDEN=1`.

## Scope honesty (amended after independent review)

- The new test/harness files are the PR's content; the branch necessarily carries the not-yet-merged #1851 base commit (dead-handler deletion), so "zero production changes" holds for this PR's own commits, not the raw branch diff until #1851 merges.
- The harness's `PageService`/registry stubs mean the goldens do **not** pin the plugin fetch machinery (`build_template_context`, shared contexts, reference extraction, widening), variable-mode collection rule evaluation, silence `mode:"page"`, or the silence cache — that list now lives in the harness docstring, and the descendant PRs carry their own targeted tests for exactly those paths.

## Non-vacuity (per the repo's test-quality bar)

Two production mutations, each breaking exactly the pinned behavior:
- dedupe check disabled → `static_page_dedupe` FAILS: `5 sends [...] vs golden 1 sends` ✔
- `_silence_state_changed` forced False → `silence_window_indicator` FAILS: boundary sends drift to poll cadence ✔

Both restored; suite deterministic across repeated runs; **79 passed** alongside the four suites it was modeled on. ruff clean.

## Discoveries now pinned

- Secondaries ARE driven under `run()` — same pass as the primary, primary first, same cadence.
- Silence end is **inclusive**: still SNOOZING at the window-end instant; content returns the second after.
- Trigger activation is pass-quantized, and `TriggerService` ran on naive wall-clock `datetime.now()` — unified onto TimeService later in this stack (#1871).


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/test/1850-engine-equivalence-harness origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

Issue #1848, opened 2026-09-06
Umbrella: one storage kernel — JsonStore (atomic write + lock + schema_version) under every store


jeffredodd:
Part of the backend rework (spec approved 2026-09-05; extends epic #1730 Phase 3).

Extends #1759 from "adopt write_json_atomic at 7 sites + add locks" to a proper
persistence kernel:

- New `src/storage/` with a `JsonStore` base class providing:
  - one `write_json_atomic()` (unique staging name, fsync, rename)
  - a `threading.RLock` per store
  - `schema_version` + ordered `MIGRATIONS` support, generalizing the
    `src/pages/storage.py` pattern that CLAUDE.md already mandates
- Rebuild on it: `SettingsService`, `PageStorage`, `ScheduleStorage`,
  `CollectionStorage`, plugin-registry state, `.system-update.json`
  (folds the #1745 fix pattern)
- The seven copy-pasted write blocks and the two regressed fixed-`.tmp`
  writers (`backup/service.py`, `config_manager.py`) all adopt it

**Zero-regression protocol:** concurrent-writer tests per store (two threads,
disjoint keys, both survive), pin-first characterization of each store's
current serialization, #1736 persistence round-trips stay green.

Absorbs/extends: #1759. Related: #1745 (landed), #1762 (lands first).

jeffredodd:
Landed on `next` in #2018: `.system-update.json` now sits on `JsonStore` (atomic write, store RLock, `schema_version` 1 with a stamping v0→v1 migration, on-disk keys unchanged and pinned by a golden). Every other store was already on the kernel. ConfigManager stays kernel-equivalent (atomic + RLock + schema_version + migrations) rather than kernel-composed by design; PERSISTENCE.md records why two of its migrations must remain deferred. Closing; ships with 9.0.

Issue #1750, opened 2026-08-24
Blocking I/O on the event loop: plugin install freezes the entire API


jeffredodd:
### Why
All 180 endpoints are `async def`; many do blocking work directly on the loop:
`POST /plugins/registry/{id}/install` runs `git clone` with a 120 s timeout
(`src/api_server.py:9697` → `sources.py:388`); `/refresh` and `/force-refresh` run
the full render + 10 s board POST + up to 120 s transition on the loop (`:3114`,
`:8500`); page preview does the full plugin fan-out (`:7532`). During any of these
the whole web UI is frozen. Four handlers already use `asyncio.to_thread` — the
pattern exists, applied inconsistently.
### What
Make blocking handlers plain `def` (Starlette threadpools them) or wrap the
blocking sections in `asyncio.to_thread`, starting with install/refresh/preview.
Add a watchdog test that the API answers `/health` while an install is in flight.
### Acceptance
Health check latency stays <100 ms during a slow mocked install.

_Part of #1730 (Phase 2)._

jeffredodd:
Verified against `next` (this is not on `main` yet; it ships when `next` lands as 9.0).

Install/update run off the loop via `asyncio.to_thread` in `src/plugins/service.py` (PR #1809, 236b148dd); the refresh/preview half landed with #1826 (9fbf8d13c). Watchdog tests: `tests/test_plugin_install_event_loop.py` (5 tests) and the AST ratchet `tests/test_no_blocking_on_event_loop.py::test_no_new_blocking_call_appears_directly_in_an_async_handler`. Note the tests assert a 0.5 s health budget rather than the 100 ms in the acceptance text; the pre-fix failure was ~5 s.

Closing as fixed on `next`.

PR #1857, opened 2026-09-06
fix(api): move refresh/preview/send blocking work off the event loop

### LINKED ISSUE #1826: Blocking I/O on the event loop: /refresh, /force-refresh, and page preview still freeze the API

jeffredodd:
Closes #1826 and completes #1750. Backend-rework **Stack 0**, final PR of the chain #1851 → #1853 → #1854 → this.

## Why

Follow-up to #1750/#1809: seven `async def` handlers still ran their blocking work inline on the single event loop — `POST /refresh`, `POST /force-refresh`, page preview, **batch** preview (N pages × full plugin fan-out in one request), current-display, `PUT /settings/active-page`, and `POST /pages/{id}/send`. During any of them the board stops being driven, the whole web UI hangs, and `/health` doesn't answer.

## What

All seven converted to the established `asyncio.to_thread` pattern (#1809), with the two rules that matter:
- `_send_with_status(...)` moves as **one whole call** — its thread-local error capture is set and read inside the same sync call.
- Multi-step handlers (`set_active_page`, `send_page`, `force_refresh`) move their blocking sequence as a single `_work()` closure; input validation stays on the loop.
- `preview_pages_batch` stays one call — its per-board-size context sharing must not be parallelized away.

**The #1809 lesson, not repeated**: moving these off-loop removes the accidental serialization of concurrent sends to the same board. Audit finding: no new lock needed — every send path here funnels through `TransitionRenderMixin.render`, which already holds the per-board `_send_lock` around the cache-check→POST→cache-write. That claim is now *pinned* by `test_concurrent_renders_on_one_client_serialize`, proven non-vacuous (neutralizing `_send_lock` → 3/4 overlapping sends). Preview-cache and settings-store writes are documented as single-assignment-safe / deferred to the storage kernel (#1848). Registry reads are safe under #1854's locks — the ordering constraint from the issue.

## Zero-regression evidence

**Fail-first** — all 7 watchdogs fail pre-fix for the intended reason:
```
E   AssertionError: /health waited 5.01s behind the batch preview — the event loop was blocked
========================= 7 failed, 1 passed ==========================
```
**Post-fix**: new suite 8/8 in 1.14s; regression set (plugin-install watchdogs, route-inventory golden, api coverage suites, board client, pause, displays, override, multi-board, all 12 silence files): **833 passed, 0 failed**. ruff clean at the CI-pinned 0.16.4.

### DIFF
--- a/src/api_server.py
+++ b/src/api_server.py
@@ -3309,8 +3309,12 @@ async def refresh_display(board_id: str | None = None, payload: dict | None = Bo
         if board_id is None:
             # Every board is driven here, so a failing secondary must surface
             # too — the wrapper aggregates across the whole pass (issue #1791).
-            sent, error = _send_with_status(
-                service, "check_and_send_active_page_with_status", "check_and_send_active_page"
+            # The pass is board network I/O, so it runs in a worker thread to
+            # keep the event loop free (#1826); _send_with_status moves as one
+            # call because its failure reason lives in a thread-local that is
+            # set and read inside the same sync call.
+            sent, error = await asyncio.to_thread(
+                _send_with_status, service, "check_and_send_active_page_with_status", "check_and_send_active_page"
             )
             if error:
                 raise HTTPException(status_code=500, detail=f"Failed to refresh display: {error}")
@@ -3328,7 +3332,10 @@ async def refresh_display(board_id: str | None = None, payload: dict | None = Bo
         if rt is None:
             raise HTTPException(status_code=503, detail=f"Board client not initialized: {board_id}")
         is_primary = board_id == get_settings_service().get_primary_board_id()
-        sent, error = _send_with_status(
+        # Board network I/O — off the event loop (#1826); _send_with_status
+        # moves as one call (thread-local failure reason, see above).
+        sent, error = await asyncio.to_thread(
+            _send_with_status,
             service,
             "check_and_send_for_board_with_status",
             "check_and_send_for_board",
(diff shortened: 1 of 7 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/fix/1826-event-loop-blocking origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

Issue #1766, opened 2026-08-24
REST conventions: response shapes, status codes, error contract, tags


jeffredodd:
### Why
Three sibling create endpoints return three shapes; zero 201s; only 19/179
endpoints declare `response_model`; ~30 return sites answer failure with HTTP 200 +
`{"success": false}`; `HTTPException.detail` is sometimes a string, sometimes a
dict; ~23 handlers take untyped `request: dict`; duplicate endpoints (3× cache, 2×
install, 3× "what's on the board"); route-shadowing on `/plugins/{id}` vs literal
segments; no versioning or tags.
### What
Write the convention doc (bare bodies, real status codes, single error shape,
response_model everywhere); apply per-domain as routers extract (#1756/#1757/#1758);
deprecate duplicates with Deprecation+Sunset headers; validate ids against
reserved route words.
### Acceptance
Extracted domains 100% response_model'd and tagged; convention doc in
docs/internal/.

_Part of #1730 (Phase 4)._

jeffredodd:
Verified against `next` (this is not on `main` yet; it ships when `next` lands as 9.0).

8cecea35f. `docs/internal/reference/API_CONVENTIONS.md` exists and `tests/test_api_conventions_ratchet.py` enforces response_model, no-200-on-failure, typed_body and declared-errors across 22 converted domains; one error shape via `src/api_errors.py`. Two small follow-ups noted but not blocking: the doc's Identifiers section claims reserved-word id validation that does not exist in code, and `/plugins/install` vs `/plugins/registry/{id}/install` are still both live. `auth` joins the ratchet in the open PR for #1926.

Closing as fixed on `next`.

PR #1858, opened 2026-09-06
refactor(web): split api.ts into per-domain modules; unify SSE auth path

### LINKED ISSUE #1763: Split web/src/lib/api.ts into per-domain modules
### Why
2,588 lines: ~165 interfaces + one ~170-method object literal. The AI chat streams
through a separate fetch path that bypasses the 401/409 redirect logic
(`api-stream.ts:54`). Known drift: `demo_plugin_id` missing from TS Page/PageCreate.
### What
Per-domain modules re-exported through the same `api` object (no call-site churn);
route the SSE stream through the shared auth-aware fetch; fix the known drift;
run /check-types.
### Acceptance
No import churn outside lib/; ts-sync-validator clean.

_Part of #1730 (Phase 3)._

jeffredodd:
Closes #1763. Backend-rework **frontend tail**, stacked on Stack 0 (#1857).

## What

`web/src/lib/api.ts` (~2,800 lines: ~165 interfaces + one 178-method object literal) is now a **62-line barrel** over 16 per-domain modules in `web/src/lib/api/` (largest 356 lines). Also:
- **SSE auth unified**: the AI chat stream's `onopen` non-ok path now routes through the shared `redirectToLoginIfNeeded` — previously it bypassed the 401/409 logic entirely.
- **Known drift fixed**: `demo_plugin_id?: string | null` added to `Page`/`PageCreate` (mirrors the Pydantic model; `PageUpdate` lacks it in Python, so TS matches).

## Zero-regression evidence

- Bodies were **sliced line-for-line by script** from the original, not retyped — a coverage check proved every non-blank line of the original landed in exactly one module.
- **Zero call-site churn, proven**: `from "@/lib/api"` import sites = 145 before and after; key-set diff of the old vs composed `api` object = identical 178 methods; export-name diff loses no symbol.
- **SSE fail-first**: the new 401-redirect test run before wiring — `Number of calls: 0` — passes after the one-line change; a companion pins that a plain 500 still does not redirect.
- node:24 container: `tsc --noEmit` clean; vitest baseline 1611 passed / 0 failed → after **1613 passed / 0 failed** (+2 new tests). prettier + eslint clean on all 19 files.
- #1825's `ApiError` / `skipAuthRedirect` / bearer-capable token functions preserved byte-for-byte in `core.ts`/`auth.ts`.

Drift scan findings (reported, intentionally not fixed): `ScheduleEntry.resolved_*` fields are route-layer additions, and `BoardStatus` has no Pydantic model (matches the inline `/status` dict) — both for the #1766/#1849 conventions pass.


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1763-api-ts-split origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1859, opened 2026-09-06
feat(web): surface per-board init errors in the UI

### LINKED ISSUE #1829: Surface per-board init errors in the UI (nothing renders GET /status boards[].error)
Follow-up from #1813 (closes #1749).

## What exists

#1813 records why each board was skipped in `DisplayService.board_init_errors` and exposes it on `GET /status` as `boards[<id>].error`, with a matching `BoardStatus.error` field in `web/src/lib/api.ts`.

## What is missing

Nothing in the web UI reads it. In fact nothing reads `status.boards` at all — the only `useStatus()` consumers are `web/src/components/service-controls.tsx` and `web/src/components/settings/about-card.tsx`, and neither touches the per-board map (which has been on `/status` since #1244).

So a user whose board is dark because of a bad credential still sees no explanation anywhere in the app. The reason exists only in the container logs and in a raw `GET /status` response.

## Why it was not done in #1813

It is not a small wiring change — there is no existing per-board health surface to hang it on. It needs a UX decision about where the warning lives (sidebar board selector badge? board settings page? a dashboard banner?), plus i18n keys across the 14 locale files and tests. That is its own PR.

## Suggested scope

- Pick one surface. The board selector in the sidebar is the strongest candidate: it is the one place that already enumerates the fleet, so an unavailable board can carry a badge there and the detail can live on that board's settings page.
- Read `status.boards[id].error`, render it next to the board, and link to that board's settings.
- Add the i18n keys to all 14 locales.

jeffredodd:
Closes #1829. Backend-rework **frontend tail**, stacked on Stack 0 (#1857). Sibling of #1858 (disjoint files; `api.ts` untouched here — verified).

## What

#1813 records why each board failed to initialize (`GET /status` → `boards[id].error`), but nothing in the UI read it — a board dark from a bad credential had no explanation anywhere. Now:

1. **Hardware settings, board card**: a destructive "Unavailable" badge on the collapsed card header (tooltip = the reason) and a destructive Alert in the expanded card with the verbatim backend reason plus a hint that saving the connection form retries initialization (which it genuinely does — every PUT re-runs `_reinitialize_board_clients()`).
2. **Dashboard banner**: one Alert per failed board, fleet-wide (the failed board is usually not the selected one), with an "Open board settings" action.

Both read the existing shared `useStatus()` query — no new polling. i18n keys added and translated across all 14 locales.

## The sidebar-selector finding

The issue's first-choice surface was a badge on the sidebar `BoardSelector` — but that component lives in `@fiestaboard/ui` and its published props (`6.0.3` d.ts) offer **no per-item adornment/status/slot of any kind**, so the wiring side can't attach one. Per the repo rule (presentation changes go in FiestaUI), this PR uses the app-owned surfaces above; the upstream prop gap is filed as a FiestaUI issue (linked below).

## Zero-regression evidence

- **Fail-first**: 6 tests written against a null-stub first — `4 failed | 2 passed` (all positive assertions failed on missing testids; the two absence tests' non-vacuity is covered by their positive twins). After: **6/6**.
- node:24 container: baseline 1611 passed / 0 failed → **1617 passed / 0 failed** (exactly the +6 new). `tsc`, prettier, eslint clean.

### DIFF
--- a/web/app/routes/home.tsx
+++ b/web/app/routes/home.tsx
@@ -13,6 +13,7 @@ import { Home as HomeIcon, Info } from "lucide-react";
 import { useEffect, useState } from "react";
 
 import { ActivePageDisplay } from "@/components/active-page-display";
+import { BoardInitErrorBanner } from "@/components/board-init-error-banner";
 import { SilenceImminentBanner } from "@/components/silence-imminent-banner";
 import { useWizard } from "@/components/wizard-provider";
 import { useTranslations } from "@/i18n/translations";
@@ -61,6 +62,7 @@ export default function Home() {
         )}
 
         <PageSection>
+          <BoardInitErrorBanner />
           <SilenceImminentBanner />
           <ActivePageDisplay />
         </PageSection>
--- a/web/messages/de.json
+++ b/web/messages/de.json
@@ -132,6 +132,9 @@
     "description": "Überwachen Sie die Anzeige Ihres Boards und die Systemaktivität",
     "noBoardConfigured": "Kein Board konfiguriert",
     "noBoardDescription": "Ihr Board ist noch nicht eingerichtet. Verbinden Sie Ihr Board, um Inhalte anzuzeigen.",
+    "boardInitErrorTitle": "{boardName} ist nicht verfügbar",
+    "boardInitErrorTitleUnnamed": "Ein Board ist nicht verfügbar",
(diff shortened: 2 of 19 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1829-board-error-surfacing origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

Issue #1759, opened 2026-08-24
One write_json_atomic() and real locks for the lock-free stores


jeffredodd:
### Why
`atomic_io.py` factors out only the staging filename; the 12-line write block is
copy-pasted 5× and the two writers that didn't copy it regressed to the fixed-.tmp
collision it prevents (`backup/service.py:236`, `config_manager.py:562`). Its
docstring claims every store holds a threading.Lock; `SettingsService`,
`PageStorage`, `ScheduleStorage`, `CollectionStorage`, and `PluginRegistry` have
none — `SettingsService` rewrites all 12 sections from memory on every mutation, so
concurrent PUTs lose each other wholesale.
### What
`write_json_atomic(path, data)` adopted at all 7 sites; RLocks on the five
lock-free stores; fix the docstring.
### Acceptance
Concurrent-writer tests per store (two threads, disjoint keys, both survive).

_Part of #1730 (Phase 3)._

jeffredodd:
Verified against `next` (this is not on `main` yet; it ships when `next` lands as 9.0).

Superseded by #1848 and delivered there: single `write_json_atomic` in `src/atomic_io.py` with unique staging names, zero raw `json.dump` in `src/`, the two regressed fixed-`.tmp` writers (backup, config_manager) adopted, and RLocks on settings/pages/schedules/collections/registry. Concurrent-writer acceptance: `tests/test_storage_kernel.py::TestConcurrentWriters`.

Closing as fixed on `next`.

Issue #1745, opened 2026-08-24
.system-update.json written non-atomically with no lock


jeffredodd:
### Why
`_system_update_state_save` uses plain truncating `open("w")`
(`src/api_server.py:1831`) and is read-modify-written concurrently from the hourly
loop and two handlers. Crash mid-write → truncated file → loader swallows the error
and silently resets the auto-update toggle. The exact #1304 bug class, missed in
one spot.
### What
Use `staging_path` (or the Phase-3 `write_json_atomic`) + a lock around
read-modify-write.
### Acceptance
Concurrency test on the state file; kill-mid-write leaves valid JSON.

_Part of #1730 (Phase 1)._

PR #1860, opened 2026-09-06
feat(storage): one JSON kernel — atomic writes, real locks, schema migrations — under every store

### LINKED ISSUE #1848: Umbrella: one storage kernel — JsonStore (atomic write + lock + schema_version) under every store

jeffredodd:
Closes #1848 (umbrella; absorbs #1759, folds the #1745 pattern). **Track A**, stacked on #1855. Nine commits, one per store, each independently green.

## Why

The 12-line atomic-write block was copy-pasted ~5×, and the writers that didn't copy it regressed to the fixed-`.tmp` collision it prevents (backup, ConfigManager's boot snapshot, plus a third found in api_server's pre-update snapshot). `atomic_io.py`'s docstring claimed every store holds a lock; **five stores had none** — `SettingsService` rewrites all 12 sections from memory on every mutation, so concurrent PUTs lost each other wholesale.

## What

- **Kernel**: `src/storage/json_store.py` — `JsonStore` with default paths via the #1762 seam, an RLock (`lock` exposed for composed critical sections), `load()/save()/mutate(fn)`, and pages-style ordered schema migrations with backup-before-first-migration. `load()` migrates in memory and lets the domain resave through its own serialization — that's what makes byte-compatibility achievable. `write_json_atomic` (fsync + rename, `private=True` for 0600) and `write_text_atomic` in `atomic_io.py`; docstring lie fixed.
- **Adopted**: pages, schedules, collections, panels (domain keeps its own `MIGRATIONS`; kernel runs them), SettingsService (all 21 mutators under the store lock — the lost-PUT fix), ConfigManager + backup + api_server snapshot + auth (atomic-write adoption). Registry verified to persist only through ConfigManager — nothing to adopt. `.system-update.json` already correct from #1745 — untouched.

## Zero-regression evidence

- **Pin-first**: golden on-disk fixtures for all five formats, generated from the pre-kernel code and verified round-trip-stable on it *before* any store was touched; byte-for-byte after.
- **Fail-first**, deterministic (event-sequenced, not barrier-raced): a PUT paused mid-dump gets lapped, then corrupts `settings.json` with its stale snapshot while its own rename ENOENT is swallowed —
```
E   json.decoder.JSONDecodeError: Extra data: line 76 column 2 (char 1535)
FAILED ...test_settings_two_concurrent_puts_to_different_sections_both_survive
```
The four storage-class concurrency tests were each proven non-vacuous by swapping the old store back in.
- **Full suite**: baseline `1 failed, 5617 passed` (known worktree-env failure) → `1 failed, 5647 passed` — identical failure set, +30 kernel tests. One test file adapted (it built `SettingsService` via `__new__`, bypassing `__init__`) with in-place justification; no assertion weakened. ruff clean.

### DIFF
--- a/src/api_server.py
+++ b/src/api_server.py
@@ -31,7 +31,7 @@ from pydantic import BaseModel, Field
 load_dotenv()
 
 from . import __version__  # noqa: E402
-from .atomic_io import write_json_atomic  # noqa: E402
+from .atomic_io import write_json_atomic, write_text_atomic  # noqa: E402
 from .auth import is_auth_enabled  # noqa: E402
 from .auth.middleware import AuthMiddleware  # noqa: E402
 from .auth.routes import router as auth_router  # noqa: E402
@@ -2311,11 +2311,10 @@ def _take_settings_snapshot(
         else:  # pragma: no cover - effectively unreachable
             logger.warning("Could not find a free snapshot filename")
             return None
-        # Use a temp file + atomic rename so a crash mid-write can't leave a
-        # truncated snapshot in place.
-        tmp = target.with_suffix(target.suffix + ".tmp")
-        tmp.write_text(document, encoding="utf-8")
-        tmp.replace(target)
+        # Atomic staged write (process-scoped staging name) so a crash
+        # mid-write can't leave a truncated snapshot, and a concurrent
+        # process can't collide on a fixed .tmp name.
+        write_text_atomic(target, document)
     except OSError:
         logger.exception("Failed to write pre-update settings snapshot")
         return None
--- a/src/atomic_io.py
+++ b/src/atomic_io.py
@@ -1,28 +1,33 @@
-"""Naming for the staging files used by FiestaBoard's atomic JSON writes.
+"""Atomic file writes for FiestaBoard's JSON stores.
 
 Every long-lived store under ``data/`` (config, settings, pages, schedules,
-collections, auth) writes by staging a temp file next to the target and then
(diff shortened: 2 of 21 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1848-storage-kernel origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1861, opened 2026-09-06
perf(engine): one shared context per tick; cached silence window

### LINKED ISSUE #1752: Change-driven render: shared context per tick, end the collection double fan-out
### Why
The loop calls `preview_page(force_refresh=True)` every tick and diffs *after* the
full render (`src/main.py:732`, `:763`). `preview_pages_batch` shares one context
across board sizes and is never used by the engine (`src/pages/service.py:557`).
Variable-mode collections build the full context twice per tick
(`src/collections/service.py:169` + `engine.py:325`). The 1 Hz silence probe
deep-copies config ~5×/second (`main.py:1137` → `config.py:577`).
### What
Build one context per tick per board size and pass it to collection resolution and
render; short-circuit render when no input changed; cache the parsed silence
window, invalidated on config write.
### Acceptance
Instrumented tick test: unchanged inputs → zero plugin fetches and zero renders;
config-lock acquisitions per second drops to ~0 at idle.

_Part of #1730 (Phase 2)._

jeffredodd:
Closes #1752. **Track B**, stacked on the equivalence harness (#1856) — which is the whole point: this removes wasted work while the golden send-sequences prove behavior is unchanged.

## The three wastes removed

1. **Per-render context builds**: every board render built its own full plugin context. Now `check_and_send_active_page` shares one context per tick per board size, built lazily by `PageService.shared_context_for` and threaded through resolution and render (new params optional — MQTT, `/refresh`, and API callers unchanged).
2. **Collection double fan-out**: variable-mode collections built the full context twice per tick (resolve + render). `resolve_page_id` now takes a lazy `context_factory`; the loop passes a closure returning the same shared context. Time/random modes never invoke it.
3. **Silence probe deep-copies**: the 1 Hz probe deep-copied config ~5×/second. `ConfigManager` now carries a `config_generation` bumped on save; the parsed silence window is cached per board keyed on it. Generation-less stub managers bypass the cache, protecting every existing test double.

No speculative render-skip was introduced: the existing content diff still gates sends, and failed/throttled sends still leave the dedupe cache clear — which is why the goldens can stay byte-identical.

## Instrumentation (fail-first — all pre-fix numbers observed, saved with the tests)

| Metric | Before | After |
|---|---|---|
| Context builds, one pass, two same-size boards | 2 | 1 |
| Context builds, variable-mode collection tick | 2 | 1 |
| Silence-window full parses, 60× 1 Hz idle probes | 120 | 1 |
| Silence migration lock-runs over those probes | 60 | 1 (+1 per config write) |

Non-vacuity by mutation: disabling context sharing kills the sharing tests; ignoring the generation kills the invalidation test.

## Zero-regression evidence

- **All 9 golden scenarios pass with zero re-records** — `tests/golden/engine/*.json` byte-identical (git-clean).
- Full suite: `5627 passed, 1 failed` — exactly the baseline failure set (known worktree-env failure). A 3-failure cluster in one serial batch was reproduced identically at HEAD (pre-existing order-dependent pollution; passes in isolation and under `-n auto`).
- Test stubs widened to accept the new optional kwargs only; no assertion weakened. ruff clean.

### DIFF
--- a/src/collections/service.py
+++ b/src/collections/service.py
@@ -106,9 +106,10 @@ class CollectionService:
         - For ``variable`` mode, walks ``variable.rules`` in order and returns
           the first ``page_id`` whose expression evaluates truthy. If
           ``context`` is None, ``context_factory`` (when given) supplies it —
-          the display loop passes its per-tick shared context this way
-          (issue #1752) so resolution and render share one plugin fan-out;
-          otherwise it is built lazily from the plugin registry.
+          the display loop passes its per-tick shared BOARD-AGNOSTIC context
+          this way (issue #1752) so every board's resolution shares one
+          ``board=None`` plugin fan-out per tick; otherwise it is built
+          lazily from the plugin registry (also board-agnostic).
         - For ``random`` mode, returns the shuffle-bag page for the current
           duration window (deterministic, stateless, no back-to-back repeats).
 
@@ -192,6 +193,11 @@ class CollectionService:
         if collection.variable is None:
             # Should be caught by validation; defend anyway.
             return collection.page_ids[0]
+
+        if not collection.variable.rules:
+            # No rules to evaluate — the default always wins; skip the
+            # plugin fan-out a context build would cost.
+            return collection.variable.default_page_id
 
         ctx = context
         if ctx is None and context_factory is not None:
--- a/src/collections/service.py
+++ b/src/collections/service.py
@@ -12,6 +12,7 @@ blank the board.
 import logging
 import math
 import time
+from collections.abc import Callable
 from datetime import datetime
 from typing import Any
 
@@ -96,6 +97,7 @@ class CollectionService:
         ref_id: str,
         now_unix: float | None = None,
         context: dict[str, Any] | None = None,
+        context_factory: Callable[[], dict[str, Any] | None] | None = None,
     ) -> str | None:
         """If *ref_id* is a collection, return the page that should be shown.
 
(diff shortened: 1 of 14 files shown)


jeffredodd:
**Post-merge measurement correction** (independent audit, measured against \`main\`):

Two characterizations in this PR and its issue are wrong and should not survive into the changelog:

1. **"Change-driven tick" is a misnomer.** The loop is still time-driven — same 1 Hz wake, same \`schedule.every(polling_interval)\`. No dirty-signal or invalidation entry point exists. What became demand-driven is the *fetch set*, not the tick. Accurate description: **time-driven tick with per-pass context sharing, demand-driven fetch, and (after #1867) off-tick sends.**
2. **The "double fan-out" is not ended, only halved.** A variable-mode collection tick still calls \`build_template_context\` **twice** (measured: 2 on both sides) — necessarily so, since rule evaluation needs a board-agnostic context and the render needs a board-aware one. What improved is fetches inside those builds: **10 → 6** per tick.

What the PR did deliver, measured over 60 idle probes with two boards: config deep-copies **6,120 → 0**, `get_feature` parses **360 → 0**, config file-lock acquisitions **600 → 0**, median probe **37.2 µs → 14.4 µs**. Context builds for two same-size boards **2 → 1**; fetches for that case **10 → 1**.

The unmet third acceptance criterion of #1752 (zero renders on unchanged input — both branches still render every tick and diff after) is tracked as #1883.

jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1752-change-driven-tick origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1862, opened 2026-09-06
perf(plugins): fetch only referenced + trigger plugins; one persistent bounded executor

### LINKED ISSUE #1751: Demand-driven plugin fetch: only fetch plugins the page references
### Why
Every render fans out to every enabled plugin whether or not the template
references it (`src/plugins/registry.py:1528`) — a page using one variable pays the
full integration fan-out every 15 s per board. The engine already parses
`{{plugin.var}}` refs. A fresh `ThreadPoolExecutor` is also built per render and
abandoned with `shutdown(wait=False)` (`:1536`/`:1568`), leaking one thread per
tick per hung plugin.
### What
Filter the fan-out to referenced plugin ids (plus trigger plugins); replace the
per-render pool with one persistent bounded executor.
### Acceptance
Test: page referencing 1 of 3 enabled plugins → exactly 1 fetch. Thread count
stable across 100 simulated ticks with a hanging plugin.

_Part of #1730 (Phase 2)._

jeffredodd:
Closes #1751. **Track B**, stacked on #1861 (chain: #1856 → #1861 → this).

## What

1. **Demand-driven fetch**: new `extract_template_plugin_ids()` scans a template's `{{...}}` roots (instance keys included, colors/specials skipped) and the fan-out fetches only those plugins — falling back to fetch-all the moment a `{{= ...}}` formula appears (owners aren't statically parseable; safety first). The per-tick shared context (#1861) records which ids it covers, so a later consumer needing more fetches **only the missing ids** and merges — variable-mode collections still see everything, nothing is fetched twice per pass.
2. **Trigger freshness is un-bypassable**: the union with trigger-capable plugins happens inside `build_template_context` itself, so no caller can starve `check_triggers` of fresh data.
3. **One persistent bounded executor** (module-level, 8 workers, lazy, shut down in the API lifespan) replaces the per-render pool. Timeout semantics preserved exactly: `futures_wait` still solely bounds the render; abandoned fetches finish harmlessly — now as **bounded occupancy instead of a leaked thread per tick per hung plugin**; saturation logs a warning.

## Zero-regression evidence

**Fail-first instrumentation** (verbatim in the test evidence):
- 1-of-3 referenced → fetches **3 → 1**
- unreferenced trigger plugin: pre-fix dropped (`{alpha,beta,trig} != {alpha,trig}`) → always fetched
- 100 ticks with a hanging plugin: **+100 threads → ≤ 10** (observed `100 <= 10` failure pre-fix)
- fetch-all fallback + cache-widening proven non-vacuous by mutation (`#REF` output / missing plugin observed, restored)

**Goldens**: all 9 engine equivalence scenarios byte-identical, zero re-records; the #1861 instrumentation suite green. Targeted suites: 887 passed across engine/templates/pages/collections/silence/triggers. Full suite: **5634 passed** ×3 consecutive clean runs (sole failure = known worktree-env image-formats test). ruff clean.

Merge-order note: `build_template_context` is also touched by #1854 (registry locks) on Stack 0 — the rebase resolution is planned (snapshot-then-release + filter compose cleanly).

### DIFF
--- a/src/api_server.py
+++ b/src/api_server.py
@@ -909,6 +909,15 @@ async def lifespan(app: FastAPI):
     except Exception:
         logger.debug("Failed to stop mDNS during shutdown", exc_info=True)
 
+    # Stop the shared plugin-fetch pool (issue #1751). wait=False: a wedged
+    # plugin fetch must not stall process shutdown.
+    try:
+        from .plugins.registry import shutdown_plugin_fetch_executor
+
+        shutdown_plugin_fetch_executor()
+    except Exception:
+        logger.debug("Failed to stop plugin-fetch executor during shutdown", exc_info=True)
+
 
 # Create FastAPI app
 app = FastAPI(
--- a/src/pages/service.py
+++ b/src/pages/service.py
@@ -419,7 +419,11 @@ class PageService:
                 return context
 
             board = board_context_for(device_type, notes_wide, notes_tall)
-            context.update(registry.build_template_context(board, plugin_ids=missing))
+            # Widening fetches EXACTLY the missing ids: the first build for
+            # this size already fetched the trigger plugins (they are in
+            # ``fetched``), so re-unioning them here would fetch each trigger
+            # plugin once per widening consumer (#1862 review).
+            context.update(registry.build_template_context(board, plugin_ids=missing, include_trigger_plugins=False))
             if plugin_ids is None:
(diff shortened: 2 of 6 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1751-demand-driven-fetch origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1863, opened 2026-09-06
refactor(api): extract pages/schedules/collections routers (pure moves, golden-gated)

jeffredodd:
Closes #1756; part of #1849. **Track A**, stacked on #1860 (sibling of the config-cleanup branch).

## What

The first three domains leave the 10k-line monolith — 32 routes across four commits:
- `test(api)`: **response-shape golden infrastructure** — 65 seeded TestClient scenarios (success + error cases) recording `{method, path, status, shape-with-values-replaced-by-types}` per domain, with a coverage assertion that the scenario set exactly matches the moved-route scope, and `RECORD_RESPONSE_GOLDEN=1` re-recording. Non-vacuity proven in-commit (a deliberate 404→400 flip and a field rename both caught).
- `refactor(api)`: **pages (16 routes)** → `src/pages/routes.py` — `/pages/ai/*` deliberately excluded (AI-domain surface sharing the AI semaphore/SSE machinery; moves with the AI domain).
- **schedules (11)** → `src/schedules/routes.py` (silence-schedule routes are settings-domain, untouched).
- **collections (5)** → `src/collections/routes.py`.

Routers are included at the exact file positions the blocks occupied, so registration order (specific-before-parameterized) is byte-identical.

## The load-bearing design decision

The suite patches domain deps as `src.api_server.<name>` (151× `get_settings_service`, 80× `get_service`, …). Moved handlers import those names **at call time from api_server** — the pattern `src/mqtt/commands.py` already uses — because a module-level import would either cycle or silently detach dozens of existing tests from their patch targets (making them vacuous). Domain-local helpers moved with their routers; shared helpers stayed. The migration of patch targets to service modules is filed as a follow-up note for #1766/#1849.

## Zero-regression evidence

- Response-shape goldens: **zero changes** after each move. Route-inventory golden (#1731): **zero edits** (records path+methods+name; all survive).
- MCP server audited: imports none of the moved handlers — no repoints.
- Per-commit domain batteries unmodified and green (783/438/158). Full suite: baseline `5647 passed / 1 failed` → `5650 passed / same 1 failed` (known worktree-env failure; +3 golden tests). Import sanity + ruff clean.

Six pre-existing findings deliberately not fixed (pure-move discipline) are listed in the final commit for the #1766 conventions pass — including `clear_page_cache` reaching a private service member and `GET /schedules?board_id=*` hardcoding `"enabled": False`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

Issue #1793, opened 2026-08-31
send-message doesn't consider line length of Note


mountainsandcode:
**Describe the bug**
Use send-message (e.g, via MQTT from Home Assistant) to a Note doesn't properly wrap the line, so the message continues into Nirvana in Row 1.

Alternatively, it would be great to see in the documentation how to force a new line in this API.

**To Reproduce**
Send a message via MQTT which is longer than the width of the board.

**Expected behavior**
Two options:
- Automatic line wrapping based on board length
- Add documentation how line wrapping can be forced (less preferred option)

**Screenshots / Logs**
If applicable, add screenshots or paste relevant log output (`docker compose logs -f`).

mountainsandcode:
@jeffredodd Awesome work on this project and thanks for tackling this (and some other tickets I raised) so quickly!

PR #1864, opened 2026-09-06
refactor(config): delete dead code, retire features.*, make env vars a real override layer

### LINKED ISSUE #1761: Delete dead code; retire features.* legacy config; make env vars a real override layer
### Why
~2,340 LOC have no production importer: `formatters/message_formatter.py` (506),
`utils/{weather,muni,surf,air_fog,star_trek_quotes}.py` (~1,640),
`security/secrets.py` (195). The dead utils are the last readers of the legacy
`features.*` config. Separately, `_apply_env_overrides` writes env values into
`features.*` AFTER the feature→plugin migration (`config_manager.py:786-955`,
order at `:389`) — on migrated installs `WEATHER_API_KEY` is a silent no-op — and
permanently mutates config.json (unsetting the env var doesn't revert).
### What
Delete the dead modules + their tests; retire the `features.*` branch of
config.py/config_manager.py; rework env handling into a read-time override that
maps to plugins.* and never persists.
### Acceptance
Env var set → live plugin config reflects it; unset → reverts; no features.* reads
remain.

_Part of #1730 (Phase 3)._

jeffredodd:
Closes #1761. **Track A**, stacked on #1860 (sibling of #1863). Net **−3,817 lines** across three green commits.

## 1. Dead code (−2,944)

Deleted with zero production importers verified: `src/utils/{weather,muni,surf,air_fog,star_trek_quotes}.py`, the whole `src/security/` package, and their tests. Two audit corrections found by verification rather than trust:
- `message_formatter.py` was **not** fully dead — `wrap_message_text` (the #1793 message-send path) uses `split_into_lines`. Kept exactly that (506→97 lines), deleted the nine dead `format_*` methods.
- Three uncalled factories inside *kept* utils were the last `features.*` readers there — pruned.

## 2. features.* retirement

Boundary established by reading consumers, not the audit list: **kept** `features.silence_schedule` (live system), the v1→v2 feature→plugin migration (upgrade path; tests unchanged), and the color-rules legacy fallback. **Retired** the 11 legacy plugin blocks, all feature classproperties, and dead helpers; live readers (muni stops, stock search, traffic validate — a hidden `getattr` key the audit missed, HA source, timezone) repointed to `plugins.*`. `Config.get_summary` keeps its exact key set.

**Bonus bug fix**: corrupt-config recovery used a shallow `DEFAULT_CONFIG.copy()`, so later in-place writes mutated the module constant itself — exposed as an order-dependent failure during the retirement, now a deep copy.

## 3. Env vars become a read-time overlay

Pre-fix, `_apply_env_overrides` wrote env values into `features.*` *after* the migration — on every migrated install `WEATHER_API_KEY` was a **silent no-op** — and permanently mutated config.json. Now: 30 env vars map to `plugins.<id>.<key>` at the single read choke point (`get_plugin_config`/`get_all_plugin_configs`), computed per call — set wins, unset reverts, **never persisted**. The two read-merge-write paths pass `include_env_overrides=False` so a masked round-trip can't persist an env secret. Board-connection env vars untouched (separate mechanism, tests confirm).

## Zero-regression evidence

- **Fail-first**: 8/11 new tests fail pre-fix for the documented reasons (silent no-op, persisted-env leak, guard greps).
- Full suite: baseline `5647 passed / 1 failed` → `5498 passed / same 1 environmental failure` (delta = deleted dead tests minus new coverage). ~15 tests that mocked retired `Config.*` attributes were repointed to the new read path with assertions preserved (e.g. muni no-key still asserts 400/500). ruff + import sanity clean.

Follow-up noted: audit `docs/` for any env-var text claiming seed-into-config behavior.


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1761-config-cleanup origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

Issue #1849, opened 2026-09-06
Umbrella: service layer everywhere — every domain gets service + router; api_server.py becomes an app factory


jeffredodd:
Part of the backend rework (spec approved 2026-09-05; extends epic #1730 Phases 3–4).

Extends #1756–#1758 from three extractions to the whole surface: one PR per
domain, each PR = record response-shape golden → extract `service.py` (where
missing) + `routes.py` → apply the #1766 conventions (bare bodies, real status
codes incl. 201s, single error shape, `response_model` on every endpoint,
OpenAPI tags, typed request models) with old shapes served under
`Deprecation` + `Sunset` headers.

Domain order: pages → schedules → collections → plugins (#1757, breaks the
api_server↔mcp_server import cycle) → system-update (#1758) → display/board
control (incl. the 5 trigger routes) → settings → auth leftovers → AI →
misc/diagnostics. Ends with `src/api_server.py` as a thin app factory.

**Zero-regression protocol:** route-inventory golden (#1731) plus a per-domain
response-shape golden recorded BEFORE each move; extraction PRs must show an
empty golden diff; convention changes alter goldens deliberately with
deprecation shims. Convention doc lands in `docs/internal/` before the first
domain.

Absorbs/extends: #1756, #1757, #1758; carries #1766 domain-by-domain.

jeffredodd:
Two latent ops-registry notes from the independent review of #1873/#1874, for the #1766 conventions/server-side-execution pass:

1. `_adapt_replace_page` hardcodes a flagship fallback — server-side `registry.execute("replace_page", ...)` on a Note-only install would create a flagship page. Latent today (nothing in production calls `registry.execute()`; chat still applies ops client-side), but it is the advertised entry point for the server-side execution migration.
2. `registry.execute()` applies chat-schema validation whenever the supplied name equals the op's chat name — and for ~12 ops the canonical name IS the chat name. The first canonical-only executor kwarg added later (e.g. `board_id` on a schedule op) would be silently dropped for canonical-name callers via `_model_fields`. The validation branch should key on which grammar the caller used, not name equality.

Also for the record: the review confirmed the doc's no-`_private` rule is already bent by mcp_server's pre-existing `_polled_characters`/`_mask_sensitive` reads — fold cleaning those into the service-layer completion.

jeffredodd:
[...]

Issue #1743, opened 2026-08-24
Plugin config writes can persist the "***" mask sentinel over real secrets


jeffredodd:
### Why
`list_installed_plugins` returns masked config; MCP `configure_plugin` merges it
back (`src/mcp_server.py:391`) and neither MCP nor `update_plugin_config`
(`src/api_server.py:8813`) strips `***` server-side. Only the web UI form defends.
An MCP client can silently destroy a stored API key.
### What
Strip mask sentinels server-side in the shared config-write path (drop keys whose
incoming value equals the mask), covered by the #1735 round-trip test.
### Acceptance
Write-masked-value-back leaves the real secret intact via both REST and MCP.

_Part of #1730 (Phase 1)._

Issue #1735, opened 2026-08-24
E2E: plugin lifecycle round-trips — fan-out isolation, secret masking, install→configure→uninstall


jeffredodd:
### Why
Plugin E2E is largely route-mocked; both uninstall specs test only the Cancel
button; no test does a secret write-`***`-read-back for plugin config (the AI
settings have exactly this test, `ai.spec.ts:222`); nothing asserts one raising
plugin doesn't block the fan-out for the rest.
### What
pytest + TestClient:
1. Two stub plugins enabled, one raises in fetch → page renders with the healthy
   plugin's variables.
2. Configure a plugin with an API key → GET returns `***` → PUT the masked value
   back → real key still stored (pairs with Phase 1 mask-sentinel fix).
3. Full install → configure → uninstall against a local fixture registry (no
   network); config purged after uninstall.
### Acceptance
All through HTTP endpoints, not service internals.

_Part of #1730 (Phase 0)._

PR #1865, opened 2026-09-06
feat(plugins): PluginService + router — one orchestration path; MCP decoupled from api_server

### LINKED ISSUE #1757: Create a plugins service + router (absorb the copy-pasted orchestration)
### Why
25 plugin endpoints drive registry + config_manager directly, copy-pasting the same
mutate→persist→reset-display→reset-engine sequence five times
(`src/api_server.py:8809-9787`), with private-member reaches
(`registry._update_status`, `config_manager._mask_sensitive`). MCP imports these
handlers from inside functions, creating the api_server↔mcp_server cycle.
### What
`src/plugins/service.py` owning the orchestration; router for the 25 endpoints;
MCP calls the service (breaking the import cycle).
### Acceptance
No `_private` access from routes; cycle gone; #1735 lifecycle tests green.

_Part of #1730 (Phase 3)._

jeffredodd:
Closes #1757; part of #1849. **Track A**, stacked on #1863. Four commits; `api_server.py` shrinks **10,257 → 9,104 lines**.

## What

- **Response-shape goldens first**: 43 records covering all 25 plugin routes incl. error branches and the masked-secret round trip; deterministic stubbed scenario driver; proven non-vacuous (a renamed response key fails it).
- **`PluginService`** (23 unit tests): every mutating plugin operation (config update, enable/disable, instances, installs, updates, uninstall) goes through one orchestration path; the mutate→persist→`reset_display_service()`+`reset_template_engine()` tail that was copy-pasted five times is written once in `reset_runtime()`. The only sanctioned private reaches (`_mask_sensitive`, `_update_status`) live behind `mask_config`/`clear_update_status` — routes have **zero** private-member access (grep-proven).
- **Plugins router**: pure move onto the service; the options endpoint's machinery deliberately stays in api_server because `test_plugin_options_route.py` monkeypatches ~20 of its module attrs — the moved handler resolves them at call time, keeping those tests non-vacuous (same patch-seam discipline as #1863).
- **MCP cycle break**: the six mutating MCP plugin tools now drive `PluginService`, not api_server handlers. Pinned by a fresh-interpreter test that exercises all six end-to-end and asserts `src.api_server` never enters `sys.modules` — proven non-vacuous by re-adding an import. (Honest framing: importing mcp_server was already lazy; the pinned improvement is that *using* plugin tools no longer loads api_server. The remaining `get_system_info`/`set_active_page` imports are other domains' work.)

## Zero-regression evidence

- Plugins goldens + route-inventory golden **byte-identical** across all four commits.
- The #1735 acceptance gate `test_plugin_lifecycle_round_trip.py` passes **unmodified**; one fixture in `test_plugin_config_mask_roundtrip.py` gained a patch line because it encoded the old MCP→REST-handler *wiring* (not the #1743 behavior) — assertions untouched, disclosed in-commit.
- Full suite: baseline `5650 passed / 1 failed` → `5675 passed / same 1 environmental failure` (+25 = exactly the new tests). ruff clean.

Findings deliberately not fixed (follow-ups): the background auto-updater still pops `_update_status` directly (non-route site — clean `PluginService` adoption later); `GET /plugins/{id}/data` still runs the plugin inline on the loop (pre-existing, flagged).


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1757-plugins-service origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

Issue #1736, opened 2026-08-24
E2E: persistence round-trips — settings survive restart, backup import/export


jeffredodd:
### Why
`test_settings_persistence` writes and re-reads in one process; nothing
re-instantiates the app against the same data dir. Backup E2E covers only the
confirm dialog and error toast. Phase 3 touches every store's on-disk shape.
### What
1. pytest: write settings/pages/schedules → tear down all singletons → rebuild app
   on the same tmp data dir → all state identical.
2. pytest: export backup → mutate state → import → state restored.
### Acceptance
Byte-level or model-level equality assertions; covers config.json + settings.json +
pages/schedules/collections.

_Part of #1730 (Phase 0)._

Issue #948, opened 2026-06-11
Integrations Lost on Upgrade


DarthXoc:
**Describe the bug**
After upgrading from 7.0.10 to 7.1.1, my integration settings seem to have been lost.

**To Reproduce**
1. Install 7.0.10
2. Configure FiestaBoard with a few Integrations
3. Upgrade to 7.1.1
5. ...
6. Observe that no integrations except 'Date & Time' are installed

**Expected behavior**
Integration settings persist across versions upgrades as they have in the past.

**Screenshots / Logs**
[app.log](https://github.com/user-attachments/files/28821017/app.log)

**Environment**
- OS: Ubuntu 24.04
- Docker version: 29.5.3
- FiestaBoard version: 7.1.1
- Board connection mode: Cloud
- Hardware: MFF PC

**Additional context**
Performed upgrade by restarting Docker. After startup and noticing that integrations disappeared, I backed up my config and tried restoring from a known previous good configuration with no luck. When that didn't work, I reapplied the configuration that I had just backed up in case this is something that Claude can fix.

jeffredodd:
Hey @DarthXoc a few questions for you. When you said you restarted docker, did you restart and pull the latest image? Or are you using the fiestaupdater to update fiestaboard via the UI? 

What has happened in the past for folks is they run their docker up command in a different repository while setting their data folder relative to that pth. That might be something to check. 

I'm also going to look into this as well to make sure we didn't have a regression. 

Edit: You can disregard this since it's confirmed this is an issue.

cordell25:
@jeffredodd - I also lost all of my integration settings on upgrading, I went straight to 7.2.1 using the update button within FiestaPi

jeffredodd:
🤔 This is weird... why didn't I? Ok, I'll give this more thought. Sorry for the trouble y'all.

DarthXoc:
Some more information for you that might help resolve this...

I set my Integrations back up again this morning and a few hours later noticed that my schedule wasn't running. I noticed under Settings > System that FiestaBoard's status was 'Stopped', so I restarted Docker. I had updated to 7.3.0 prior to setting my Integrations back up again, so there was no upgrade performed. When FiestaBoard came back up, all of my Integrations had been reset again.

In looking at the `config.json` on my host system, it looks it was overwritten with the default `config.json`.

I imported my latest backup, watched `config.json` get updated, then stopped/started Docker again and watched it again get overwritten with the default.

PR #1866, opened 2026-09-06
feat(settings): unify board credentials on settings.json

### LINKED ISSUE #1760: Unify board credentials on settings.json
### Why
Credentials live in `config.json → board.*` AND `settings.json → boards[]`, synced
once at first boot then diverging (`settings/service.py:859`). Five legacy readers
still use the config.json copy (`api_server.py:3500/7070/7186/7234`,
`main.py:319`). Structural root of the #948/#1102 "board went offline after I
changed a key" family.
### What
Schema-versioned migration of the legacy block into settings.json; repoint the
five readers; delete the copy-once seam; keep `GET/PUT /config/board` as a shim
over settings during deprecation.
### Acceptance
#1736 persistence tests; upgrade test from a legacy config.json fixture.

_Part of #1730 (Phase 3)._

jeffredodd:
Closes #1760. **Track A**, stacked on #1864 (chain: #1855 → #1860 → #1864 → this).

## Why

Board credentials lived in `config.json → board.*` AND `settings.json → boards[]`, synced once at first boot then diverging — the structural root of the #948/#1102 "board went offline after I changed a key" family. Six legacy readers (the audit predicted five; `Config.get_summary` was a sixth) still read the stale config.json copy.

## What

- **Schema-versioned settings migration (v2→v3)**, on the #1848 kernel machinery: a fully credential-less primary board inherits the legacy `config.json` block; a board with **any** credential is never touched (settings is the maintained copy — precedence proven idempotent). The copy-once seam is deleted; its sole descendant is a first-boot seed when settings has no boards at all. Exactly one function still reads the legacy block: `_read_legacy_board_connection()`, used only by the migration.
- **Six readers repointed** to the settings store (welcome-message send, system-info, connection-info fallback, network diagnostics, `_build_board_clients`' legacy fallback — removed outright so a stale config key can't produce a ghost client — and `Config.get_summary`). `GET /config/validate` deliberately keeps its dual-source boolean first-run probe (documented).
- **`GET/PUT /config/board` is now a deprecation shim**: shapes pinned by golden-style tests, `Deprecation` + successor `Link` headers (existing repo pattern), PUT writes the primary settings board only (masked secrets preserved) and provably does **not** touch config.json. The on-disk legacy block stays as a rollback copy for older versions — never read at runtime.
- Env-var seeding (`BOARD_READ_WRITE_KEY` et al.) unchanged; it reaches settings via the migration/first-boot seed — pinned by a dedicated test and the CI env-var flow.

## Zero-regression evidence

- **Fail-first**: 8/18 new tests fail pre-fix on the exact bug shapes (stale key B served while settings has key A; ghost legacy client; PUT persisting to config.json), plus a revert-to-HEAD non-vacuity proof for the divergence test.
- `tests/test_persistence_round_trip.py` (#1736 gate) green **unmodified**, including under `BOARD_READ_WRITE_KEY=test_key`.
- Full suite: baseline `5498 passed / 1 failed` → `5517 passed / same 1 environmental failure`. Storage golden: the only byte changed is `schema_version` 2→3 (the intentional format change; the credentialed golden board untouched). Old-contract tests re-encoded with renames that say what changed (e.g. `..._falls_back_to_legacy_config` → `..._never_reads_legacy_config`). ruff clean.

(The base branch got a CI fix in the same push: the workflow's verify-imports step hardcoded two modules #1761 deleted.)

### DIFF
--- a/src/api_server.py
+++ b/src/api_server.py
@@ -3768,9 +3768,7 @@ async def send_welcome_message():
         logger.error(f"Failed to create board client: {e}")
         raise HTTPException(status_code=503, detail=f"Board not configured: {str(e)}") from e
     if board_client is None:
-        raise HTTPException(
-            status_code=503, detail="Board not configured: no board with a usable connection"
-        )
+        raise HTTPException(status_code=503, detail="Board not configured: no board with a usable connection")
     board_client.skip_unchanged = False  # Always send the welcome message
 
     try:
@@ -4034,6 +4032,7 @@ async def validate_config():
     # first-run. Board-related validation errors/missing_fields from the
     # legacy config are dropped in that case.
     has_configured_board_instance = False
+    has_connection_attempt = False
     try:
         from .devices import BoardInstance
 
@@ -4043,11 +4042,23 @@ async def validate_config():
                 instance = BoardInstance.from_dict(b)
             except Exception:  # pragma: no cover - defensive
                 continue
+            if instance.has_connection_attempt:
+                has_connection_attempt = True
             if instance.is_connection_configured:
                 has_configured_board_instance = True
                 break
     except Exception:  # pragma: no cover - defensive
         logger.exception("Failed to inspect multi-board settings during validate_config")
+
+    # A board with SOME connection detail but not a working set is
+    # *misconfigured*, not first-run: it must surface as a per-board error
+    # (#1813), never bounce an existing install back into the setup wizard.
+    # Before #1760 the legacy config.json copy masked this case; with
+    # settings as the single credential source the distinction is load-
+    # bearing (a token-less note array replacing the only board used to
+    # keep the wizard away purely via the stale legacy copy).
+    if has_connection_attempt and not has_configured_board_instance:
+        is_first_run = False
 
     if has_configured_board_instance:
         is_first_run = False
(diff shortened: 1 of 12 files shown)


jeffredodd:
E2E root cause found and fixed in b70dcc27e: the e2e helper seeds credentials via the legacy `PUT /config/board`, which now (correctly) writes settings only — so when the two note-array specs replaced the fleet with a token-less note array, `validate_config` saw *no* configured board anywhere and flipped the app into the setup wizard, which is what actually hid the Settings heading. The fix distinguishes **misconfigured** from **never configured**: any explicitly-entered connection detail (`BoardInstance.has_connection_attempt`) keeps `is_first_run` false, and incomplete credentials surface as per-board errors (#1813/#1829's surface) instead of re-triggering onboarding. Fail-first tests cover both new behaviors plus the preserved blank-board and DELETE-reset wizard paths. This also fixes a real upgrade-path regression, not just the test: saving your only board with incomplete credentials would have bounced a working install into the wizard.

jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1760-board-credentials origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1867, opened 2026-09-06
feat(engine): per-board send workers — transitions no longer stall the fleet

### LINKED ISSUE #1755: Run transitions off the tick thread
### Why
Transitions run synchronously on the tick thread under the send lock, up to 120 s
(`src/transitions/runner.py:20`, `src/main.py:789`) — one board's animation stalls
the other boards, the silence detector, and collection rotation.
### What
Per-board send worker (thread) so a transition blocks only its own board; tick
enqueues instead of sending inline.
### Acceptance
Fake-clock test (#1734 harness): board 2 updates on schedule while board 1 runs a
long transition.

_Part of #1730 (Phase 2)._

jeffredodd:
Closes #1755. **Track B**, stacked on #1862 (chain: #1856 → #1861 → #1862 → this). The highest-risk engine change, landed under the golden corpus.

## Design

- **`BoardSendWorker`** (new `src/displays/send_worker.py`): one **latest-wins** queue per board runtime (at most one pending job), drained by a demand-scoped daemon thread. The tick computes everything as before and enqueues only the final `client.render` + its post-send bookkeeping. Per-board serialization is untouched — every send still funnels through the client's `_send_lock`.
- **The `*_with_status` contract survives**: send paths grew keyword-only `wait: bool = True`. Every API/MQTT/test caller waits (bounded 150s > max transition) and gets the real outcome — the thread-local error capture is snapshotted into the completion closure, so `/refresh` still reports its own pass's failure. Only the engine loop passes `wait=False` — which is the whole point: board 2 keeps its cadence while board 1 runs a 120s transition.
- **Preemption immediacy preserved**: enqueue sets the client's `_cancel_transition` event — mirroring `render()`'s own first move — so a running transition winds down when a newer frame is *enqueued*, not when dequeued.
- **In-flight dedupe** (necessary addition, disclosed): fire-and-forget ticks skip when the identical send key is already queued/executing, else every tick during a transition would replay the animation on landing. `wait=True` callers bypass it (refresh must preempt + re-send). Proven by mutation.
- Runtime-swap epoch guard (`runtimes.get(id) is rt`) prevents a superseded worker writing stale state; rebuilds join workers with a bounded 1s timeout so a wedged send can't stall them; `DisplayService.wait_until_idle()` is the public drain primitive.

## Golden discipline

- Scenarios 1–5, 7–9: **byte-identical**, verified both by comparison and by re-recording producing zero git diff; **30 consecutive suite runs green** (determinism via `wait_until_idle` before every simulated-time advance).
- Scenario 6 (two boards): golden file **unchanged**; the sanctioned cross-board relaxation landed as sort-normalization *plus* a new unsorted per-board-subsequence assertion — strictly stronger than a re-record.
- **New scenario 10** (`slow_transition_no_cross_board_stall`) is the issue's acceptance test. Fail-first, pre-change: *"board-1's long transition finished before run() completed: the tick thread sat inside the send"* — the run only ended because a watchdog released it. Post: board 2 pinned at full 15s cadence while board 1 blocks; latest-wins drain asserted (frames 2–4 provably never sent).

## Validation

718 targeted tests green (equivalence, timing, silence ×7, multi-board, per-board, MQTT, both instrumentation suites, API); full suite failure set **identical to baseline** (known environmental failure only; one intermittent config-migration episode reproduced on pristine HEAD — pre-existing pollution, not this change). Three post-implementation mutations (guard off, adoption off, cancel-signal off) each failed exactly their covering test. ruff clean.


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1755-per-board-send-workers origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

Issue #1794, opened 2026-08-31
Active page exposed via MQTT doesn't change after send-message, can't restore said page


mountainsandcode:
**Describe the bug**
The active page exposed via MQTT (e.g., to Home Assistant) always stays on the last selected page, even if the content of the board has changed (e.g., externally or via send-message API).

While the web application in Fiestaboard notices this ("Board was changed by another app"), the MQTT select remains static, which also means that I cannot restore the last page.

**To Reproduce**
1. Select a page
2. Send-message
3. Try to select the same page in Home Assistant (via MQTT)

**Expected behavior**
I would like to be able to restore properly via MQTT (either via the select in Home Assistant, or by pressing MQTT button "Refresh Display" which also does nothing), and for the Active page to properly reflect that there is a non-page selected

PR #1868, opened 2026-09-06
feat(board-client): connection retries with backoff; universal per-type send floor

### LINKED ISSUE #1754: Board client: retries, backoff, and a universal send-interval floor
### Why
Every send is a single POST, timeout 10 s, no retry (`src/board_client.py:563`) —
an unreachable board burns 10 s of the display thread per board per tick. Rate
limiting exists only for note arrays, via an unlocked module-global racing
concurrent senders (`:79`, `:521-532`); RW Cloud has no client-side floor at all.
### What
Small retry/backoff policy; per-board-type min-send-interval; per-board lock
around throttle state; shorter connect timeout for LAN boards.
### Acceptance
Unit tests for backoff schedule and floor; #1732 integration stays green.

_Part of #1730 (Phase 2)._

jeffredodd:
Closes #1754. **Track B**, stacked on #1867 (chain: #1856 → #1861 → #1862 → #1867 → this).

## What

- **Retry/backoff**: `_post_with_retry` — 2 attempts max, retrying only connection-level failures (the board never answered); HTTP 4xx/5xx are answers and never retry. Backoff waits on the client's cancel event, so a preempting frame abandons the retry immediately. Worst case stays far under #1867's 150s wait bound.
- **Universal send floor, per board type**: note array 15s (existing constant, exact semantics and log format preserved — all 6 existing throttle tests pass unmodified), **RW Cloud 15s new** (matching the documented limit in `docs/setup/cloud-api.md`), local unfloored. `_admit_send` does floor → unchanged-content → slot reservation atomically under a per-instance lock *before* the POST, with rollback on failure so a failed send doesn't burn the window; the `last_send_throttled` → engine dedupe-cache contract (#1794) is preserved byte-for-byte. `min_send_interval_ms` now reports the per-type floor so the transition runner paces cloud frames.
- **Timeout split**: connect tightened to 3s LAN / 5s cloud; read stays 10s.

Deliberate deviation, justified: the note-array throttle store stays module-level (token-keyed) because cross-client-recreation persistence is an explicitly tested contract — it gains a module lock; the new cloud floor is per-instance as specced.

## Zero-regression evidence

- **Fail-first: 14/19 new tests fail pre-fix**, including deterministic reproductions of both target bugs — cloud double-send inside the window, and the unlocked note-array race (thread held inside the mocked POST past the throttle check; pre-fix both POSTs go out). The 3 pre-fix-passing tests were proven non-vacuous by three mutations, each killed by exactly one named test.
- **Goldens byte-identical**, zero re-records; engine/send-worker/safety-net batch 176 green; board-client batteries 141 green.
- Full suite: strict subset of baseline (all remaining failures pre-existing pollution-class, present in baseline, none board-client related — they're the exact class Track A's #1855 eliminates on merge). ruff clean.

### DIFF
--- a/src/api_server.py
+++ b/src/api_server.py
@@ -3558,6 +3558,42 @@ async def get_board_current_message(force: bool = False, board_id: str | None = 
     }
 
 
+def _throttled_send_response(board_client) -> JSONResponse | None:
+    """A 429 for an out-of-band write dropped by the client-side send floor.
+
+    #1868 review: cloud boards (and note arrays) enforce a minimum interval
+    between sends (#1754); a send inside the window returns ``(True, False)``
+    with ``last_send_throttled`` set — the content was DROPPED, not
+    delivered, and unlike the engine tick (which retries next pass) the
+    manual out-of-band endpoints (/send-message, /send-welcome-message,
+    /debug/blank, /debug/fill, /debug/info) never retry. Answering
+    "success/unchanged" would silently swallow the user's write, so they
+    answer 429 with a Retry-After hint computed from the floor.
+
+    Returns None when the last send was not throttled (the ``is True`` guard
+    also keeps Mock clients in tests, whose attributes are truthy, on the
+    legacy path unless they opt in).
+    """
+    if getattr(board_client, "last_send_throttled", False) is not True:
+        return None
+    try:
+        floor_ms = int(getattr(board_client, "min_send_interval_ms", 0))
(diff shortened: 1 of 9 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1754-board-client-policy origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1869, opened 2026-09-06
refactor(api): extract the system-update subsystem into src/system/ (pure move)

### LINKED ISSUE #1758: Extract the system-update subsystem into src/system/
### Why
~700 lines of snapshot handling, version comparison (Docker Hub + GitHub), and file
I/O live inline in route handlers (`src/api_server.py:1725-2415`); `src/system/`
exists but holds only https_certs and mdns.
### What
`src/system/update_service.py` (+ storage for the state file, building on #1745);
routes become thin.
### Acceptance
Pure move verified by existing system-endpoint tests + golden route file.

_Part of #1730 (Phase 3)._

jeffredodd:
Closes #1758; part of #1849. **Track A**, stacked on #1865 (chain: #1855 → #1860 → #1863 → #1865 → this).

## What

~760 lines of update-check, snapshot, and updater-sidecar machinery leave the monolith:
- **`src/system/update_service.py`** — Docker Hub + GitHub version comparison (the #1430 newest-of-both-sources behavior intact), the fiestaupdater client, the settings-snapshot family (retention 5, filename allow-list), and the update state machine with the #1745 RLock + atomic-write semantics byte-identical. One new function: `run_system_update_check_if_due()` — the hourly tick body, so the lifespan loop shell stays in api_server and calls the service (the issue's suggested split).
- **`src/system/models.py`** — the 9 request/response models, verbatim. **`src/system/routes.py`** — 8 thin handlers (`/version`, `/system/update-check|update/status|update|update/rollback|update/auto|restart|shutdown`), same endpoint function names so the route-inventory golden is untouched.
- Boot-time post-upgrade restore family deliberately stays in api_server: it runs from lifespan and its tests monkeypatch `api_server.<name>` around direct module calls — moving it would churn fixtures for zero behavior gain.

## Zero-regression evidence

- Response-shape goldens (21 records) recorded **pre-move**, zero-diff after; non-vacuity proven in-commit (deleting a response field → `response drifted`).
- `tests/test_system_endpoints.py` passes **unmodified**; every `patch("src.api_server.<name>")` seam kept live (re-import block + call-time resolution; the state-file/snapshot-dir constants stay on api_server and the service reads them back through it, so existing monkeypatches still steer the service).
- Full suite: baseline `5859 passed / 1 failed` → `5860 / same 1 environmental failure`. ruff at the CI-pinned version clean — including a small format fix to the #1865 golden commit that would otherwise have failed the pinned `ruff format --check` stack-wide (cherry-picked to #1865's branch as well).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -170,6 +170,8 @@ ignore = [
 # api_server's lint-ignore set so the extraction stays a pure move; the
 # convention cleanup happens with #1766.
 "src/pages/routes.py" = ["PTH", "SIM", "RET", "PIE", "RUF010", "RUF013", "RUF059", "RUF100"]
+"src/system/routes.py" = ["PTH", "SIM", "RET", "PIE", "RUF010", "RUF013", "RUF059", "RUF100"]
+"src/system/update_service.py" = ["PTH", "SIM", "RET", "PIE", "RUF010", "RUF013", "RUF059", "RUF100"]
 "src/schedules/routes.py" = ["PTH", "SIM", "RET", "PIE", "RUF010", "RUF013", "RUF059", "RUF100"]
 "src/collections/routes.py" = ["PTH", "SIM", "RET", "PIE", "RUF010", "RUF013", "RUF059", "RUF100"]
 # CodeQL-sensitive files: see the matching [tool.ruff.format].exclude entry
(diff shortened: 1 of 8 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1758-system-update-service origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1870, opened 2026-09-06
feat(engine): backoff retry for boards that failed to initialize

### LINKED ISSUE #1827: Periodically retry boards that failed to build a client at startup
Follow-up from #1813 (closes #1749).

#1813 isolates per-board init: a board whose client cannot be built is skipped with its reason recorded in `DisplayService.board_init_errors`, and the rest of the fleet keeps running.

## Gap

There is no periodic retry. A board that failed for a *transient* reason — the board was unplugged, DHCP had not handed out its address yet, the Local API was still booting — stays down until something calls `rebuild_board_clients()`.

Today the only triggers for that are the board-settings mutation paths, i.e. the user re-saving the board's settings. Recovery works, but it needs a human, and the human has no reason to suspect a save is what fixes it.

## Suggested shape

- A low-frequency retry (order of minutes) that re-attempts only the boards present in `board_init_errors`, not a full `rebuild_board_clients()`.
- Backoff, so a permanently misconfigured board does not log an error every cycle for the life of the process.
- Must not race the send path: `_build_board_clients` swaps `self.runtimes`, and a retry landing mid-render is a new failure mode. Reusing the existing poll thread rather than adding another is preferable — note that `initialize()` is now idempotent w.r.t. that thread.

## Not in #1813 because

It is a behaviour change with its own design surface (cadence, backoff, thread ownership, interaction with in-flight sends) rather than part of the isolation fix, and it needs its own tests.

jeffredodd:
Closes #1827. **Track B**, stacked on #1868.

## What

#1813 isolated per-board init failures but left recovery manual — a board that failed for a transient reason (unplugged, DHCP late, Local API booting) stayed down until a human re-saved its settings. Now the existing poll thread (no new thread) runs `_retry_failed_board_inits()`: only boards in `board_init_errors` are re-attempted (never a full rebuild), with 60s backoff doubling to a 900s cap. Log discipline per the issue: debug per attempt, error only when the backoff step increases, info on recovery — a permanently misconfigured board does not spam the log for the life of the process.

Race safety: recovery swaps runtimes via the same atomic copy-insert-replace the rebuild uses, stops a replaced worker with the bounded join, and a new `_runtimes_lock` guards the two writers' read-modify-write swaps (the rebuild path had no synchronization to reuse — two concurrent swaps could lose each other's updates); readers stay lock-free. A board-settings rebuild clears the retry schedule so a fresh save restarts at 60s.

## Zero-regression evidence

- **Fail-first: 4/4 new tests fail pre-fix** (no retry exists; zero attempts recorded; method absent). Backoff test pins the exact schedule (60/180/420/900/1800/2700/3600 over a simulated hour) *and* asserts error-log count == backoff steps, not poll cycles.
- All 10 engine goldens byte-identical (the harness never starts the poll thread; recovery is pinned at poll-loop/method level on the fake clock). No-race test: recovery lands while another board's worker is blocked mid-send.
- Full suite: strict subset of baseline (known environmental + pollution-class flakes only, verified present at baseline). ruff clean.

Known pre-existing limitation, documented not changed: if *every* board fails at startup, `initialize()` still returns False before the poll thread exists.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -244,14 +244,26 @@ class DisplayService:
         return bid if bid else self._PRIMARY_FALLBACK_KEY
 
     def _ensure_primary_runtime(self) -> BoardRuntime:
-        """Return the primary runtime, creating an empty one if needed."""
+        """Return the primary runtime, creating an empty one if needed.
+
+        The placeholder insert runs under ``_runtimes_lock`` with a
+        double-checked read (#1870 review): unlocked, the insert raced the
+        recovery pass's copy-swap of ``self.runtimes`` — the placeholder
+        could land in the pre-swap dict (and be lost), or land after the
+        swap and clobber the freshly recovered runtime whose init error was
+        already cleared, silently re-stranding the board. The common case
+        (runtime already exists) stays lock-free.
+        """
         rt = self._primary_runtime()
         if rt is None:
             key = self._resolve_primary_key()
-            rt = self.runtimes.get(key)
+            rt = self.runtimes.get(key)  # unlocked fast path
             if rt is None:
-                rt = BoardRuntime(client=None, board_id=key)
-                self.runtimes[key] = rt
+                with self._runtimes_lock:
+                    rt = self.runtimes.get(key)  # double-check under the lock
+                    if rt is None:
+                        rt = BoardRuntime(client=None, board_id=key)
+                        self.runtimes[key] = rt
             self._primary_board_id = key
         return rt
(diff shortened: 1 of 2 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1827-board-recovery origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

Issue #1767, opened 2026-08-24
Triggers: persist them or delete them


jeffredodd:
### Why
`TriggerService` is pure in-memory (all active triggers lost on restart), and the
5-route `/triggers` REST family has zero frontend consumers. It's either an
unfinished feature or a fossil.
### What
Decide: (a) finish — add storage with schema_version + UI surface, or (b) remove
the routes and service. Audit recommends deciding before #1756 touches the routes.
### Acceptance
Either persisted-with-tests or fully deleted; no half-state.

_Part of #1730 (Phase 4)._

jeffredodd:
Decision (backend-rework design review, 2026-09-05): **keep + integrate**, not persist-or-delete as framed.

The audit framing was half-wrong: triggers are a live plugin-facing feature, not a fossil. The display loop evaluates trigger-capable plugins every tick (`src/main.py:_check_trigger_override`), the published **calendar-sub** plugin declares `supports_triggers: true`, and `plugins/_template` documents the contract for plugin authors. Deleting `TriggerService` would break an external plugin and a documented API.

What actually holds from the issue:
- "Lost on restart" is mostly benign — active triggers are re-derived from plugin state every tick. What restart really loses is **dismissals** (a dismissed trigger comes back).
- The 5 REST routes are an unused-by-frontend ops surface over the live service, not a parallel feature.

Disposition, tracked in the engine-redesign umbrella #1850:
- Keep the plugin contract (`supports_triggers`, `check_triggers()`, `trigger_page_id`) unchanged.
- Trigger evaluation becomes a first-class input stage of the redesigned tick pipeline.
- Persist **dismissals only**, schema-versioned, on the storage kernel (#1848).
- The 5 REST routes ride the display-domain router extraction (#1849).

PR #1871, opened 2026-09-06
feat(triggers): TimeService clock + persisted dismissals; trigger check as a named engine stage

jeffredodd:
Part of #1850 (final Track B item); implements the #1767 disposition. Stacked on #1870 — the full Track B chain: #1856 → #1861 → #1862 → #1867 → #1868 → #1870 → this.

## What

The plugin trigger contract (`supports_triggers`, `check_triggers()`, `trigger_page_id`) is untouched — calendar-sub keeps working. Three gaps closed:

1. **Clock unification**: `TriggerService` ran on module-level naive `datetime.now()` — outside the app's clock discipline entirely (discovered building the harness, which had to patch it as a fifth clock seam). All trigger timing now goes through `get_time_service()`, aware-UTC throughout. **The fifth seam is deleted from the harness** — and the fail-first evidence proves it was load-bearing: with the seam removed pre-fix, golden scenario 7 diverged (the trigger never expired on wall clock); post-fix the golden is byte-identical with no seam.
2. **Persisted dismissals** — the one real restart loss: a dismissed trigger came back after restart. Suppressions now persist (`schema_version: 1`, atomic writes, pruned on load/write); semantics are exactly today's, just durable. The A-track merge adapter is a single documented seam (`_default_dismissals_file()` → `get_data_dir()` + JsonStore, no migration needed — the payload is already schema-versioned).
3. **Named engine stage**: the tick's trigger check is now a documented "trigger input" stage; verified `check_triggers()` reads plugin-cached data that #1862's fetch-union keeps fresh — no behavior change, now it's written down.

## Zero-regression evidence

- **Fail-first: 8 failures pre-fix** — 7 new tests (persistence, clock) plus the golden divergence above. Non-vacuity probe: disabling the write-through fails the restart-survival test behaviorally.
- `git diff tests/golden/` empty across both commits; equivalence suite (10 scenarios), send-worker, timing, silence, and trigger suites all green; full suite failure set identical to baseline. Trigger stores in tests are now tmp-isolated (no repo `data/` pollution). ruff clean.

One disclosed shape change: `ActiveTrigger.to_dict()["activated_at"]` now carries a `+00:00` offset (aware-UTC consistency); nothing pins the naive form.

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -1303,6 +1303,11 @@ class DisplayService:
         feature set (locked epic decision); silence is resolved and delivered
         per board (issue #1788). All state reads/writes go through ``rt``.
 
+        The pass runs a fixed sequence of stages: silence evaluation, the
+        pause short-circuit, the output-target short-circuit, the TRIGGER
+        INPUT stage (#1850 — see the stage comment below), the temporary
+        override, then schedule/manual page resolution and the send.
+
         ``contexts`` is the pass-wide shared template-context cache (issue
         #1752): one dict per ``check_and_send_active_page`` pass, keyed by
         board size, so N boards of one size cost one plugin fan-out per tick.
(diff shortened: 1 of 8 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1850-triggers-integrated origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1872, opened 2026-09-06
docs(api): REST conventions reference (#1766)

jeffredodd:
First deliverable of #1766, stacked on #1869. The convention document the audit called for, landed **before** any endpoint changes shape — per-domain application follows as each remaining domain extracts (#1849), using the golden-gated mechanics the doc itself specifies (record shapes → pure move → deliberate conventions commit with deprecation shims).

Codifies: bare bodies, 201s for creates, failures never 200, one `detail` error contract, `response_model`/typed request models everywhere, router+service layout with no `_private` reaches, Deprecation/Sunset/successor-Link policy for every reachable surface, and reserved-word id validation for the `/plugins/{id}`-class shadowing bugs.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -84,6 +84,7 @@
     "ollama",
     "organise",
     "pylint",
+    "Pydantic",
     "pytest",
     "QLOCKTWO",
     "rescan",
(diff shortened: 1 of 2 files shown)


jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/docs/1766-api-conventions origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1873, opened 2026-09-06
feat(ops): one tool grammar — chat and MCP collapse onto a shared operation layer

### LINKED ISSUE #1764: One tool grammar: collapse chat_ops and MCP onto a shared operation layer
### Why
`src/ai/chat_ops.py` defines a second, parallel grammar of 19 ops with different
names for the same actions as the 28 MCP tools (`update_plugin_config` vs
`configure_plugin`, `replace_page` vs `create_page`), executed client-side by the
web UI. Teaching text (board dimensions, template syntax) is written twice and the
MCP copy is hardcoded and already stale.
### What
One service-backed operation set both surfaces call; generate teaching text from
`get_dimensions()`; one test suite covering both entry points.
### Acceptance
Behavior parity test: same op through chat and MCP yields identical state.

_Part of #1730 (Phase 4)._

jeffredodd:
Closes #1764; part of #1849. **Track A**, stacked on #1869 (sibling of #1872, the conventions doc).

## What

`src/ai/chat_ops.py` defined a second, parallel grammar of 19 ops with different names for the same actions as the 28 MCP tools, and the teaching text (dimensions, template syntax) was written twice with the MCP copy hardcoded. Now:

- **`src/ops/`**: a 24-operation canonical registry — 19 server executors calling the domain services, 5 client-side chat ops registered grammar-only. Both spellings resolve (`update_plugin_config` ↔ `configure_plugin`, `replace_page` ↔ `create_page`); chat args validate against the exact `parse_tool_call` schemas. `chat_ops.py` is a byte-compatible re-export shim over `src/ops/grammar.py`; MCP tools dispatch into the same executors.
- **Teaching text generated once** (`src/ops/teaching.py`) from live device/engine metadata, consumed by both the chat system prompt and MCP prompts.

## The divergences the unification fixed (fail-first, all three observed pre-fix)

1. **Plugin config**: chat/REST *replaced*, MCP *merged* — a partial second write via chat dropped `station_id` and 400'd. Canonical: merge (the pinned MCP contract).
2. **Interval-only collection update** force-flipped a variable-mode collection to `time` on the chat path. Canonical: mode preserved.
3. **Partial schedule update** through MCP **wiped `end_time`** (explicit `None`s defeating `exclude_unset` — the same defect class `update_page` had already fixed). Canonical: only-supplied-fields.

Stale teaching text fixed by generation: filters `|upper`/`|lower` **never existed** (real ones were missing), color range was off by one (63–70), and the formula roster was frozen at 15 of the live 50 functions — all now generated and drift-locked by tests.

## Zero-regression evidence

- Existing AI/MCP suites pass **unmodified**: `test_ai_op_contract`, `tests/ai/*`, `test_mcp_state_effects`, `test_mcp_prompts_and_resources`, `test_mcp_server`, `test_mcp_skill_quality` (413→439 with the 26 new ops tests). One deliberate wiring-test update: the ≥33-service-call vacuity floor split across mcp_server (22) + executors (11) with in-place justification — total unchanged.
- **Parity suite** (10 cross-grammar tests) locks the ops with no divergence; non-vacuity proven by deliberately re-diverging the chat adapter.
- Full suite: baseline `5689 passed / 1 failed` → `5726 passed / same 1 environmental failure`. Goldens untouched; ruff + import sanity clean.

Noted for the #1766 follow-up: the web drawer still executes chat ops client-side via REST, so the browser path keeps replace-semantics until execution moves onto the server op layer.

### DIFF
--- a/src/ai/chat_ops.py
+++ b/src/ai/chat_ops.py
@@ -1,548 +1,128 @@
-"""Tool grammar for the streaming AI chat.
+"""Compatibility surface for the chat-op grammar (issue #1764).
 
-The chat lets a model emit one or more ``fiestaboard`` fenced JSON blocks
-inline with its prose. Each block is a structured *operation* that the
-editor applies to the page in flight.
-
-We deliberately do **not** use a provider's native tool/function-calling
-API: many OpenAI-compatible servers (Ollama, LM Studio, older OpenRouter
-routes) silently ignore the ``tools`` field, and the Anthropic adapter
-in :mod:`src.ai.protocols` doesn't yet plumb ``tool_use`` content blocks
-through streaming. A fenced JSON convention works identically across
-every provider we support.
-
-The server validates each block against the schemas below as the fence
-closes, and re-emits the validated payload as an ``event: tool_call``
-SSE frame to the frontend.
+The grammar itself — the pydantic arg models, the op-name registry, and
+``parse_tool_call`` — moved verbatim to :mod:`src.ops.grammar`, next to
+the operation registry (:mod:`src.ops.registry`) that maps every chat op
+onto its canonical executor. This module re-exports the whole surface so
+existing importers (``src.ai.chat``, the contract tests, and anything
+external) keep working unchanged; what the model emits and what the web
+client validates against is byte-identical.
 """
 
 from __future__ import annotations
 
-from typing import Any, Literal, Union
-
-from pydantic import BaseModel, ConfigDict, Field
-
-Alignment = Literal["left", "center", "right"]
-
-
-# ---------------------------------------------------------------------------
-# Line-metadata block, shared by replace_page and the line-level patch ops.
-# ---------------------------------------------------------------------------
-
-
-class LineMetadata(BaseModel):
-    model_config = ConfigDict(extra="ignore")
-
-    alignment: Alignment = "left"
-    wrap: bool = False
-
-
-# ---------------------------------------------------------------------------
-# replace_page — wholesale page replacement (same shape as the legacy
-# ``/pages/ai/generate`` response). Used when the user asks for a brand-new
-# page or a full rewrite.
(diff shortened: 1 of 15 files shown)


jeffredodd:
**Post-merge audit correction** — this PR's summary overstates how wired the ops layer is. For the record before merge:

- **\`registry.execute()\` / \`execute_sync()\` have zero production callers.** Their only users are \`tests/test_op_parity.py\` and \`tests/test_ops_registry.py\`. As shipped, \`src/ops/registry.py\` is a **test-enforced equivalence harness**, not a live dispatcher.
- **MCP** delegates to \`ops.executors\` directly (18 tools) — real sharing, but it bypasses the registry.
- **Chat** uses \`src/ops/\` for *parsing only*; \`src/ai/chat_ops.py\` is a re-export shim over \`ops/grammar.py\`. Execution is still a 14-branch \`if (call.op === …)\` chain in the browser calling REST directly (\`git diff main..HEAD -- web/\` is **empty**), with the two grammars kept in sync by a comment in \`ai-chat-types.ts\`.

So "one tool grammar" is accurate about the *grammar* and about MCP's executors; it is not yet true that both surfaces dispatch through one layer. The three divergence fixes and the generated teaching text are real and delivered — that part stands.

Also flagged by the audit, worth addressing before this merges: \`src/plugins/service.py\` imports \`fastapi.HTTPException\` with 25 raise sites (every other service in the tree has zero), and \`ops/executors.py:774\` imports a *route handler* (\`from src.system.routes import system_update_apply\`) — both invert the layering this stack is establishing.

jeffredodd:

Verified rather than assumed: `git merge-base --is-ancestor origin/feat/1764-operation-layer origin/next` passes, so every commit on this branch is reachable from `next` at `3d77b6c01`.

PR #1963, opened 2026-09-12
chore(docs): regenerate app captures

roblesi:
Serialised app screens regenerated because the UI changed.

- Generated from `2c7d24dbe0fed129ab36aed67d97bce8e61117c9` on `main`
- 39 capture files

**Check the commit above is still recent before merging.** This branch is rebuilt from whatever `main` was when the run started, so a PR left open across other merges will roll them back. Re-run `Docs: regenerate app captures` to refresh it.

Generated by `.github/workflows/docs-captures.yml`; see `assets/docs-captures/README.md`. Merging syncs these to the site through `publish-docs.yml`.


jeffredodd (approved):
Approved as part of clearing the open-PR backlog. CI green on every required lane.

PR #2000, opened 2026-09-14
feat(marketplace): sort plugins by when they were added

### LINKED ISSUE #1999: Marketplace: sort plugins by when they were added
Reported on Discord by Yogesh Mhatre (running the Home Assistant add-on):

> Ability to sort plugins by release/added date? I'm running via home assistant addon, I can sort by Name, but not by newness

## Today

The Marketplace list view sorts by **Name**, **Category** and **Author** (`marketplaceSort` in `web/app/routes/integrations._index.tsx`). There is no way to see what is new, which matters more as the catalog grows — it went from 54 to 63 entries in a single day recently, and those nine are only findable if you already know their names.

## Why it isn't just a UI change

Nothing in the data carries a date. `plugin-registry.json` entries hold `id`, `name`, `description`, `repository`, `author`, `fiestaboard_version`, `icon`, `category`, and `RegistryEntry` / `get_registry_entries()` mirror exactly those. So the sort key has to be introduced first.

Options considered:

| Source | Verdict |
|---|---|
| GitHub `created_at` / `pushed_at` per repo | Needs a network call per plugin, rate-limited, and useless on a box that is offline or behind a proxy. Also measures repo creation, not catalog availability. |
| Position in the registry file | Entries are appended today, but nothing enforces it and a re-sort would silently scramble the order. Too fragile to build a visible feature on. |
| An explicit `added` date on each entry | Deterministic, offline, and honest about what it means: the day the plugin became installable. |

Going with an explicit field.

## Proposal

1. Add `"added": "YYYY-MM-DD"` to every entry in `plugin-registry.json`. The backfill is not guesswork — walking the git history of that file gives the first commit in which each id appears, and that resolves a date for all 63 current entries.
2. Carry it through `RegistryEntry` and `get_registry_entries()` so the API exposes it.
3. Have `scripts/extract_plugin.py` stamp it when it registers a new plugin, so it stays current without anyone remembering.
4. Add an **Added** column to the Marketplace list view, sortable, defaulting to newest-first when selected.

Entries added in the same bulk extraction share a date, which is accurate rather than a limitation. The two big clusters are 24 entries on 2026-03-22 and 25 on 2026-05-02.

## Out of scope

Sorting the Installed tab, and any "New" badge or filter. Worth doing separately if the column proves useful.

roblesi:
## Description

Closes #1999. Requested on Discord by Yogesh Mhatre, running the Home Assistant add-on:

> Ability to sort plugins by release/added date? I'm running via home assistant addon, I can sort by Name, but not by newness

The Marketplace list view sorted by Name, Category and Author, so there was no way to see what is new. The catalogue went from 54 to 63 entries in a day, and those nine were findable only if you already knew their names.

## Why this isn't a UI-only change

Nothing in the data carried a date. Registry entries held `id`, `name`, `description`, `repository`, `author`, `fiestaboard_version`, `icon` and `category`, and `RegistryEntry` / `get_registry_entries()` mirrored exactly that. The sort key had to be introduced first.

| Source | Verdict |
|---|---|
| GitHub `created_at` per repo | A network call per plugin, rate-limited and useless on an offline box. Also measures repo creation, not catalogue availability. |
| Position in `plugin-registry.json` | Entries are appended today, but `extract_plugin.py` re-sorts the array by `id` — the order is not a fact to build a visible feature on. |
| An explicit `added` date | Deterministic, offline, and honest about what it means. |

## What changed

- **`plugin-registry.json`** — every entry gains `"added": "YYYY-MM-DD"`. The backfill is not guesswork: the first commit in which each id appears in that file's history resolves a date for all 63 entries. Purely additive, no entry reordered.
- **`RegistryEntry` + `get_registry_entries()`** — carry and publish the field. Defaults to `""` so an entry predating it still loads.
- **`scripts/extract_plugin.py`** — stamps the date on publish, and passes the existing date back in on re-publish. Without that the entry is rebuilt from scratch and a plugin re-published years later would jump to the top of a newest-first sort.
- **Marketplace list view** — a sortable **Added** column, localised into all 14 locales.

## Ordering decisions

Each has a test:

- **First click is newest-first.** That is what someone clicking a date column is asking for. The text columns still open A-Z.
- **Same-day entries fall back to name.** Bulk extractions publish many at once, so ties are common and the order should be stable rather than incidental.
- **Entries with no date sort last in both directions.** An unknown date is not "the oldest", and letting `""` sort first would bury the newest plugins under it.

Two clusters share a date — 23 entries on `2026-03-22` and 25 on `2026-05-02`, the original bulk extractions. That is accurate rather than a limitation.

> Correction: an earlier revision of this description and of the commit message said 24 for the March cluster. The history contains a 24th id dated that day, `baywheels`, which is no longer in the registry, so the count among current entries is 23. Caught by comparing against #2001.

## Out of scope

Sorting the Installed tab, and any "New" badge or filter. Worth doing separately if the column proves useful.

## Tests

- 25 frontend tests across the 4 `integrations-*` suites, including 8 new ones covering every ordering rule above
- 353 backend tests across `test_extract_plugin`, `test_plugin_sources`, `test_plugin_registry` and `test_api_coverage`, including new ones asserting every shipped entry has a well-formed ISO date
- `eslint`, `prettier --check` and `tsc --noEmit` clean

## Relationship to #2001

#2001 is the `claude-issue-triage` workflow's automatic first pass at #1999, opened as a draft a minute after the issue was filed. It reaches largely the same design independently. This PR is the one that has been run against CI, rebased onto the merged #2002 preview-seed fix, and had its auto-review feedback triaged. See the comparison comment below.

## Related Issue

Closes #1999

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] New plugin
- [ ] Documentation update
- [ ] Refactor / maintenance

## Checklist

- [x] I have tested my changes locally
- [x] I have added/updated tests where applicable
- [x] I have updated documentation where applicable
- [x] My code follows the project's coding standards


roblesi:
Went through the five auto-reviews. All returned no blockers; here is what I did with each point.

**Fixed — docs field list.** The docs review caught that step 3 of "Submitting to the Registry" still listed the old field set while the example JSON directly below it already showed `added`. Corrected in 1fc6154.

**Filed as follow-up — sortable header a11y.** Both a11y reviews flagged that the sortable `<TableHead>` elements are `onClick`-only: no keyboard operation and no `aria-sort`, so the sort controls are mouse-only and the sort state is not announced. It is pre-existing and both reviews said not to gate this PR on it. It also affects the Installed tab, not just the Marketplace, so fixing it column-by-column here would be the wrong shape. Opened #2003 to do both tables in one pass, including the keyboard test that axe cannot catch.

**No change — date locale.** One review noted `toLocaleDateString(undefined, …)` formats to the browser locale rather than the active next-intl locale. Accurate, but every other date in the app does the same (`page-grid-selector.tsx:308`, `use-format-preferences.tsx:28-29`), and `useLocale()` is currently used only by the language selector. Making this one cell differ would be inconsistent; switching all of them is a separate decision. Left as-is deliberately.

**No change — em dash placeholder.** Flagged as reading literally as "em dash" to a screen reader, and marked low priority in the review itself. It matches the muted no-value convention used elsewhere.

CI was fully green on the rebase before this docs commit: Platform Tests passing, 28 checks, zero failures.

roblesi:
## Overlap with #2001

#2001 is not a duplicate anyone opened by hand. `claude-issue-triage.yml` runs on `issues: [opened, labeled]` and takes an automatic first pass at every new issue, opening a draft on `claude/issue-<n>`. Filing #1999 triggered it, and it produced its own implementation a minute later while this branch was already in progress.

The two arrived at largely the same design independently — same `added` field, same git-history backfill, same sortable column, same newest-first-on-first-click behaviour. That convergence is reassuring about the approach. Three differences worth recording before either is merged:

**Undated entries under oldest-first.** #2001 sorts them last only under newest-first; its own comment says so (`"" < any real date, so the desc comparison pushes them to the bottom`). Under oldest-first they sort to the top, because `""` compares before every real date. This PR pushes them last in both directions and has a test for it. No entry lacks a date today, so it is latent either way — but a third-party entry submitted without the field is exactly when it would surface, and landing at the top of an oldest-first list is a confusing place for "unknown".

**Docs.** #2001 changes no documentation. This PR documents the field in both `docs/development/plugin-guide.md` and `docs/internal/development/PLUGIN_DEVELOPMENT.md`, including the rule that re-publishing preserves the original date.

**March cluster count.** #2001 says 23 and is right; earlier revisions of this PR said 24. The history contains a 24th id dated `2026-03-22`, `baywheels`, which has since left the registry, so 23 is the count among current entries. Corrected in the description above. Credit where it is due — the discrepancy only surfaced by comparing the two.

Also worth noting this PR is the one that has actually been through CI: it surfaced the missing preview seeds (fixed in #2002), has been rebased on that fix, and is green with 28 checks. #2001 predates #2002, so its branch still carries the latent failure.

Recommendation: keep this one and close #2001, after someone confirms the two points above are judged the same way.

PR #2054, opened 2026-09-24
DO NOT MERGE: CI validation for the main-into-next merge (b0c272448)

jeffredodd:
**Do not merge. Do not review. This PR exists only to run CI, and will be closed as soon as it reports.**

`next` is not in `ci.yml`'s `push` trigger (`branches: [main, develop]`), so the
full suite never runs on a push to `next` — only CodeQL does. The
`main` → `next` merge at `b0c272448` therefore landed without Platform Tests,
UI Tests, E2E, a11y or the integration smoke ever seeing it.

This branch points at that exact merge commit so `pull_request` fires and the
suite runs against the merged tree. Targeting `main` rather than `next` on
purpose: `paths-filter` diffs against `main`, which is what makes every job
actually run instead of being skipped as unchanged.

Verified locally before the merge was pushed, for reference:

- 1927 web unit tests, 172 files
- 7809 Python tests (4 failures, all `git ls-files` / `ls-remote` exit 128
  inside the container — environmental, and identical on pre-merge content)
- `tsc --noEmit`, `ruff format --check`, `ruff check`, `prettier --check`,
  `eslint` all clean
- `plugin-registry.json` parses; no registry entry lost from either side
- Beta image published successfully for linux/amd64 and linux/arm64

The gap this closes is E2E, the a11y runs and the integration smoke, none of
which run locally.


jeffredodd:
Full CI green on the merge content: 30 pass, 4 skipped, 0 failures — including E2E, AI + MCP E2E, Auth E2E, HA Ingress E2E, both Accessibility runs, Integration smoke, Platform Tests, UI Tests and both arch builds. That was the gap: a push to `next` gets only CodeQL, since `next` is not in ci.yml's push trigger. Closing as intended; nothing here is meant to merge.

